PR #6, opened 2022-01-24
'Solution'

cloudprogramm:


Krimat (approved):

app/cinema/hall.py
```diff
@@ -0,0 +1,17 @@
...
+class CinemaHall:
+
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(self, movie_name: str, customers: list, cleaning_staff):
+        print(f"\"{movie_name}\" started in hall number {self.number}.")
+
+        for customer in customers:
+            Customer.watch_movie(customer, movie_name)
```
MasterpieceElbow, line 14: Of course you can call `Customer.watch_movie` and pass customer instance there. But it doesn't work if you pass object, that has `watch_movie` method but it isn't `Customer` instances, for example cinema staff gathered to watch movie. It is bad practice to hardcode class here. Better use `customer.watch_movie` and don't import these classes

app/cinema/hall.py
```diff
@@ -0,0 +1,17 @@
...
+        print(f"\"{movie_name}\" ended.")
+
+        Cleaner.clean_hall(cleaning_staff, self.number)
```
MasterpieceElbow, line 17: The same as above

app/main.py
```diff
@@ -1,6 +1,26 @@
...
+        customers: list,
+        hall_number: int,
+        cleaner: str,
+        movie: str
+):
+    customers_list = []
+
+    for people in customers:
+        customer = Customer(people["name"], people["food"])
+        cinema_bar = CinemaBar()
```
MasterpieceElbow, line 17: You make 4 different cinema bars here

app/people/customer.py
```diff
@@ -0,0 +1,7 @@
+class Customer:
+    def __init__(self, name: str, food: str):
+        self.name = name
+        self.food = food
+
+    def watch_movie(self, movie: str) -> None:
+        print(f"{self.name} is watching \"{movie}\".")
```
MasterpieceElbow, line 7: You don't use `'` inside, so it is better to wrap message with `'` and don't use `\"` inside

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):

PR #9, opened 2022-01-24
'Solution'

liuda-chuzdiuk:


alix-mak (approved):
I think it's good solution

app/cinema/hall.py
```diff
@@ -0,0 +1,13 @@
+from app.people.cinema_staff import Cleaner
+
+
+class CinemaHall:
+    def __init__(self, number):
+        self.number = number
+
+    def movie_session(self, movie_name, customers, cleaning_staff=Cleaner):
```
MasterpieceElbow, line 8: You set by default cleaning_staff to Cleaner class. You don't need here to set by default and import Cleaner

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):

PR #12, opened 2022-01-24
'Solution'

alexandra-halytska:


liuda-chuzdiuk (approved):

app/main.py
```diff
@@ -1,6 +1,14 @@
...
 def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+    for customer in customers:
+        CinemaBar.sell_product(Customer(customer["name"],
+                                        customer["food"]),
+                               Customer(customer["name"],
+                                        customer["food"]).food)
+    cleaner_guy = Cleaner(cleaner)
+    return CinemaHall(hall_number).movie_session(movie, customers, cleaner_guy)
```
MasterpieceElbow, line 14: You don't need to return anything here

app/cinema/hall.py
```diff
@@ -0,0 +1,21 @@
...
+
+    def movie_session(self, movie_name, customers, cleaning_staff):
+        print(f"\"{movie_name}\" started in hall number {self.number}.")
+        for customer in customers:
+            if not isinstance(customer, Customer):
+                Customer.watch_movie(Customer(customer["name"],
+                                              customer["food"]),
+                                     movie_name)
+            else:
+                Customer.watch_movie(customer, movie_name)
```
MasterpieceElbow, line 17: Don't check for instance and don't hardcode `Customer.watch_movie(customer)` here, this function suggests that customer in customers has `watch_movie` method and may not be a `Customer` instance, so it is better to use just `customer.watch_movie`

app/cinema/hall.py
```diff
@@ -0,0 +1,21 @@
...
+        print(f"\"{movie_name}\" ended.")
+        if not isinstance(cleaning_staff, Cleaner):
```
MasterpieceElbow, line 19: The same as with `Customer`

app/main.py
```diff
@@ -1,6 +1,14 @@
...
 
 
 def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+    for customer in customers:
+        CinemaBar.sell_product(Customer(customer["name"],
+                                        customer["food"]),
+                               Customer(customer["name"],
+                                        customer["food"]).food)
```
MasterpieceElbow, line 12: Just create `Customer`'s separately and pass instances to `sell_product` and  `movie_session`. It is very complicated, that you passed already created instances to `sell_product` and pass raw data to `movie_session`.

MasterpieceElbow (changes requested):

app/main.py
```diff
@@ -1,6 +1,16 @@
...
+from app.cinema.hall import CinemaHall
 
 
 def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+    listOfInstances = []
+    for customer in customers:
+        listOfInstances.append(Customer(customer["name"], customer["food"]))
+        CinemaBar.sell_product(Customer(customer["name"],
```
MasterpieceElbow, line 11: Here you created 3 `Customer` instances of the same customer. Wouldn't it be better to create `customer = Customer(...)` and use this customer everywhere?

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):

PR #13, opened 2022-01-24
Solution

phantom-terminal:


IgorOlifirenko (approved):

app/cinema/hall.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class CinemaHall:
+    def __init__(self, number):
+        self.number = number
+
+    def movie_session(self, movie_name, customers, cleaning_staff=Cleaner):
+        print(f'"{movie_name}" started in hall number {self.number}.')
+        for customer in customers:
+            Customer.watch_movie(customer, movie_name)
```
MasterpieceElbow, line 12: Don't hardcode class `Customer` here. Customers can have method `watch_movie` but they may not be `Customer` instances. Better to use `customer.watch_movie`

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):

PR #17, opened 2022-01-24
Solution

Geksus:


Oleksandr563487953 (approved):
Files with tests seem to be changed, instead of files with code. For instance, string "cinema bar sold popcorn to Bob" was changed in tests instead of bar.py file. Is it ok?

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):

PR #21, opened 2022-01-25
'Solution'

Oleksandr563487953:


app/main.py
```diff
@@ -1,6 +1,13 @@
...
 
 
 def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+    customers_instances = [
+        Customer(customer["name"], customer["food"]) for customer in customers
+    ]
+    for customer in customers_instances:
+        CinemaBar.sell_product(customer=customer, product=customer.food)
```
Abnormaltype, line 12: You have created 4 different CinemaBar()

Abnormaltype (changes requested):

app/main.py
Oleksandr563487953: No, CinemaBar class cann`t create instances, according to task description it doesn`t have a constructor. It only has staticmethod, which can be called as many times as 'for' loop iterates, and just returns a string.

app/cinema/hall.py
```diff
@@ -0,0 +1,10 @@
+class CinemaHall:
+    def __init__(self, number):
+        self.number = number
+
+    def movie_session(self, movie_name, customers, cleaning_staff):
+        print(f"\"{movie_name}\" started in hall number {self.number}.")
+        for customer in customers:
+            customer.watch_movie(movie_name)
+        print(f"\"{movie_name}\" ended.")
```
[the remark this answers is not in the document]
Oleksandr563487953: I would prefer to use f-strings in only one way, with one type of quotations, to avoid confusions in future code.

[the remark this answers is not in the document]
Abnormaltype: Okay

app/main.py
Abnormaltype: yes. it's true, overlooked

Abnormaltype (approved):

MasterpieceElbow (approved):

PR #22, opened 2022-01-25
'Solution'

serhii56465:


app/cinema/hall.py
```diff
@@ -0,0 +1,11 @@
+
+class CinemaHall:
+    def __init__(self, number):
+        self.number = number
+
+    def movie_session(self, movie_name, customers, cleaning_staff):
+        print(f'"{movie_name}" started in hall number {self.number}.')
+        for i in customers:
```
Oleksandr563487953, line 8: It would be better to use more specific cycle variable instead of 'i', it will be more meaningful and readable.

Oleksandr563487953 (changes requested):

MasterpieceElbow (approved):

PR #24, opened 2022-03-19
Solution, task with classes in differents packages and moduls. Katiukha

KatiukhaO:


app/cinema/hall.py
```diff
@@ -0,0 +1,11 @@
+class CinemaHall():
+
+    def __init__(self, number):
+        self.number = number
+
+    def movie_session(self, movie_name, customers, cleaning_staff):
+        print(f"\"{movie_name}\" started in hall number {self.number}.")
```
Magus6784, line 7: Annotation = love
And i think better wat for this case:
print(f'"{movie_name}" started in hall number {self.number}.')

app/main.py
```diff
@@ -1,6 +1,19 @@
...
 
 
 def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+    customers = [Customer(person["name"], person["food"])
+                 for person in customers]
+    hall = CinemaHall(hall_number)
+    bar = CinemaBar()
+    cleaner_worker = Cleaner(cleaner)
```
Magus6784, line 12: You create new name for cleaner (cleaner_worker) but use previous name for customers.
Just thinking this is a little bit strange

Magus6784 (approved):

app/cinema/hall.py
MasterpieceElbow, line 7: I agree with Max, use `f'  '` if you use `"` inside.

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):

PR #27, opened 2022-05-16
Solution

YuriiKindrat:


Magus6784 (approved):
I realy sure that adding annotation very good fot this case(all case), because when you are call function from another file you can see which type your variable

app/cinema/hall.py
```diff
@@ -0,0 +1,11 @@
+class CinemaHall:
+
+    def __init__(self, number):
+        self.number = number
+
+    def movie_session(self, movie_name, customers, cleaning_staff):
+        print(f"\"{movie_name}\" started in hall number {self.number}.")
```
MasterpieceElbow, line 7: You do not have to use an escape symbol `\` for a `"` in an f-string with a single quote `'` like you did in `Customer.watch_movie`.

app/cinema/hall.py
```diff
@@ -0,0 +1,11 @@
+class CinemaHall:
+
+    def __init__(self, number):
+        self.number = number
+
+    def movie_session(self, movie_name, customers, cleaning_staff):
+        print(f"\"{movie_name}\" started in hall number {self.number}.")
+        for x in customers:
```
MasterpieceElbow, line 8: `x` is a bad loop variable name.

app/main.py
```diff
@@ -1,6 +1,17 @@
...
+from app.cinema.hall import CinemaHall
+from app.cinema.bar import CinemaBar
+from app.people.cinema_staff import Cleaner
+from app.people.customer import Customer
 
 
 def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+    ls_of_customers = []
```
MasterpieceElbow, line 8: `ls` is not informative, better rename to `customers_list` or just `customers`.

app/main.py
```diff
@@ -1,6 +1,17 @@
...
+    for customer in customers:
+        ls_of_customers.append(Customer(customer['name'],
+                                        customer['food']))
+    hall = CinemaHall(hall_number)
+    bar = CinemaBar()
+    cleaner = Cleaner(cleaner)
+    for peop in ls_of_customers:
```
MasterpieceElbow, line 15: `peop` is not a good loop variable name.

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):

PR #28, opened 2022-05-18
'Solution'

DmytroSvirsa:


app/main.py
```diff
@@ -1,6 +1,15 @@
...
+from app.people.customer import Customer
+from app.people.cinema_staff import Cleaner
 
 
 def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+    humans = [Customer(customer["name"], customer["food"])
+              for customer in customers]
+    cb = CinemaBar()
```
Magus6784, line 10: I think name "bar" better than "cb"

Magus6784 (approved):
I realy sure that adding annotation very good fot this case(all case), because when you are call function from another file you can see which type your variable

app/main.py
TPolina: agree)

TPolina (approved):
Well done!

PR #29, opened 2022-05-18
Solution

4ndyua:


KatiukhaO (approved):
nice code!

app/cinema/hall.py
```diff
@@ -0,0 +1,10 @@
+class CinemaHall:
+    def __init__(self, number):
+        self.number = number
+
+    def movie_session(self, movie_name, customers, cleaning_staff):
+        print(f'\"{movie_name}\" started in hall number {self.number}.')
```
MasterpieceElbow, line 6: You should not place an escape symbol `\` for `"` if you use an f-string with single quotes `f'  '`

MasterpieceElbow (changes requested):

PR #33, opened 2022-05-19
Solution

Magus6784:


app/cinema/hall.py
```diff
@@ -0,0 +1,18 @@
+class CinemaHall:
+
+    def __init__(self, number: int):
+        self.number = number
+
+    def movie_session(self,
+                      movie_name: str,
+                      customers: list,
+                      cleaning_stuff,
+                      ):
```
YuriiKindrat, line 10: it's better to write both brackets next to the argument as in $6 or both brackets as in line $10

YuriiKindrat (approved):

MasterpieceElbow (approved):

PR #36, opened 2022-05-19
'Solution'

Aleksei-Isaev:


KatiukhaO (changes requested):
nice code, but on I think, in $8 and $18 lines, name of  variables can be more informative. For exemple,  "c" change on "customer", then you code will be readed better.
And, from $9 to $12 line , can change bilding list , create list comprehentions)

KatiukhaO (approved):

app/cinema/hall.py
```diff
@@ -0,0 +1,14 @@
+class CinemaHall:
+
+    def __init__(self, number):
+        self.number = number
+
+    def movie_session(self, movie_name, customers, cleaning_staff):
+        print(f"\"{movie_name}\" started in hall number {self.number}.")
```
MasterpieceElbow, line 7: Use `f'   '` if you need to use `"` inside.

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):

PR #39, opened 2022-05-20
'cinema_solution'

CheshkoSergey:


app/cinema/hall.py
```diff
@@ -0,0 +1,11 @@
+class CinemaHall:
+
+    def __init__(self, number):
+        self.number = number
+
+    def movie_session(self, movie_name, customers, cleaning_staff):
+        print(f'\"{movie_name}\" started in hall number {self.number}.')
```
MasterpieceElbow, line 7: You should not use an escape symbol `\` for a `"` in an f-string with a single quote `'`. `f'"some string"'` also works.

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):

PR #41, opened 2022-05-20
Develop

BohdanKryven:


app/main.py
```diff
@@ -1,6 +1,19 @@
...
-    pass
+    customers_obj = []
+    for customer in customers:
+        customers_obj.append(Customer(name=customer["name"],
+                                      food=customer["food"]))
+
+    for customer in customers_obj:
+        CinemaBar.sell_product(product=customer.food, customer=customer)
+
+    clean = Cleaner(cleaner)
```
lenasharabura, line 16: i think, it's bad naming
you can use cleaner, it will be better

lenasharabura (changes requested):
Good joob!
But i was able to find a few errors

app/cinema/hall.py
```diff
@@ -0,0 +1,20 @@
...
+        self.number = number
+
+    def movie_session(self,
+                      movie_name: str,
+                      customers: list,
+                      cleaning_staff: Cleaner):
+        print(f"\"{movie_name}\" started in hall number {self.number}.")
+
+        for customer in customers:
+            Customer.watch_movie(customer, movie_name)
```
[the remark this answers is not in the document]
BohdanKryven: Thank you for your comment! Fixed it. It's really better to use instance instead Class

app/main.py
BohdanKryven: Thanks. Renamed

lenasharabura (approved):
Good job

app/cinema/hall.py
```diff
@@ -0,0 +1,19 @@
...
+
+class CinemaHall:
+    def __init__(self, number):
+        self.number = number
+
+    def movie_session(self,
+                      movie_name: str,
+                      customers: list,
+                      cleaning_staff: Cleaner):
+        print(f"\"{movie_name}\" started in hall number {self.number}.")
```
MasterpieceElbow, line 12: If you need to use `"` inside a string - better use f-string with single quotes: `f'some string'` then you should not use escape symbol `\` inside a string.

MasterpieceElbow (changes requested):

app/cinema/hall.py
BohdanKryven: I fixed it.

MasterpieceElbow (approved):

PR #43, opened 2022-05-20
Solution

Hectorovich:


app/cinema/hall.py
```diff
@@ -0,0 +1,13 @@
+class CinemaHall:
+    def __init__(self, number):
+        self.number = number
+
+    def movie_session(self, movie_name, customers, cleaning_staff):
+        print(f"\"{movie_name}\" started in hall number {self.number}.")
```
katevasilenko, line 6: I'm not sure you need to fix it, but you can you single quotes here to avoid use **\**
f'"{movie_name}"...'
here and all code below

katevasilenko (approved):

katevasilenko (approved):

app/cinema/hall.py
MasterpieceElbow, line 6: Use `f’ “Some string” ‘` if you need to use `”` inside a string instead of `f” \” Some string \” “`

MasterpieceElbow (changes requested):

app/cinema/hall.py
MasterpieceElbow, line 6: You still have bad style here.

app/people/customer.py
```diff
@@ -0,0 +1,7 @@
+class Customer:
+    def __init__(self, name, food):
+        self.name = name
+        self.food = food
+
+    def watch_movie(self, movie):
+        print(f"{self.name} is watching \"{movie}\".")
```
MasterpieceElbow, line 7: And here.

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):

PR #45, opened 2022-05-20
Cinema-Visit Solution

mens0:


app/cinema/hall.py
```diff
@@ -0,0 +1,12 @@
+class CinemaHall:
+    def __init__(self, number):
+        self.number = number
+
+    def movie_session(self, movie_name, customers, cleaning_staff):
+        print(f"\"{movie_name}\" started in hall number {self.number}.")
```
katevasilenko, line 6: i'm not sure you need to fix it, but you can use single quotes here: **f'"{movie_name}"...'**

app/cinema/hall.py
```diff
@@ -0,0 +1,12 @@
...
+
+        for customer in customers:
+            customer.watch_movie(movie_name)
+
+        print(f"\"{movie_name}\" ended.")
```
katevasilenko, line 11: single quotes here too

app/people/customer.py
```diff
@@ -0,0 +1,7 @@
+class Customer:
+    def __init__(self, name, food):
+        self.name = name
+        self.food = food
+
+    def watch_movie(self, movie):
+        print(f"{self.name} is watching \"{movie}\".")
```
katevasilenko, line 7: single quotes here too

katevasilenko (approved):

app/cinema/hall.py
MasterpieceElbow, line 6: I agree with Kate, use `f'  '` if you need to use `"` inside.

MasterpieceElbow (changes requested):

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):

PR #48, opened 2022-05-20
Solution

NOSSUDAK:


app/cinema/hall.py
```diff
@@ -0,0 +1,15 @@
...
+    def __init__(self, number: int):
+        self.number = number
+
+    def movie_session(
+            self,
+            movie_name: str,
+            customers: list,
+            cleaning_staff
+    ) -> None:
+        print(f"\"{movie_name}\" started in hall number {self.number}.")
```
katevasilenko, line 11: I'm not sure you need to fix it, but you can write this way: **f'"{movie_name}\"...'** with single quotes

app/cinema/hall.py
```diff
@@ -0,0 +1,15 @@
...
+        for customer in customers:
+            customer.watch_movie(movie_name)
+        print(f"\"{movie_name}\" ended.")
```
katevasilenko, line 14: single quotes here too

app/people/customer.py
```diff
@@ -0,0 +1,7 @@
+class Customer:
+    def __init__(self, name: str, food: str):
+        self.name = name
+        self.food = food
+
+    def watch_movie(self, movie: str) -> None:
+        print(f"{self.name} is watching \"{movie}\".")
```
katevasilenko, line 7: single quotes here too

katevasilenko (changes requested):

app/cinema/hall.py
NOSSUDAK: In reference they used double quotes, so I did it in a same way

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):

PR #50, opened 2022-05-20
'Solution'

dashaprotsko:


app/cinema/bar.py
```diff
@@ -0,0 +1,7 @@
+from app.people.customer import Customer
```
Ed12816, line 1: this import seems superfluous, it seems to me that it would be better to do without it

app/cinema/hall.py
```diff
@@ -0,0 +1,21 @@
+from app.people.customer import Customer
+from app.people.cinema_staff import Cleaner
+
```
Ed12816, line 3: I don't think there is a need to use this.

tests/test_imports.py
```diff
@@ -1,7 +1,7 @@
 import ast
 from _ast import ImportFrom
 
-from app import main
+from app.cinema import main
```
Ed12816, line 4: change the code in the tests,  it was possible????

Ed12816 (approved):
I will approve this because  it works and passes the tests, but some  things look weird...)

app/cinema/main.py
```diff
@@ -0,0 +1,24 @@
...
+        customers_list.append(Customer(name=customer["name"],
+                                       food=customer["food"]))
+
+    for customer in customers_list:
+        CinemaBar.sell_product(product=customer.food, customer=customer)
+
+    clean = Cleaner(cleaner)
+    cinema_hall = CinemaHall(hall_number)
+
+    cinema_hall.movie_session(movie, customers_list, clean)
```
MasterpieceElbow, line 24: You were supposed to write this code inside `app/main` instead of `app/cineam/main`, rewrite.

tests/test_imports.py
MasterpieceElbow, line 4: Don't edit test file

MasterpieceElbow (changes requested):

PR #52, opened 2022-05-21
Solution

vasylhnatiuk:


lenasharabura (approved):

app/cinema/hall.py
```diff
@@ -0,0 +1,11 @@
+class CinemaHall:
+
+    def __init__(self, number):
+        self.number = number
+
+    def movie_session(self, movie_name, customers, cleaning_staff):
+        print(f'''"{movie_name}" started in hall number {self.number}.''')
```
MasterpieceElbow, line 7: Use triple quotes only for the multiline string. Use just `f'  '` here.

MasterpieceElbow (approved):

PR #53, opened 2022-05-21
Solution

Chardona:


app/cinema/hall.py
```diff
@@ -0,0 +1,18 @@
...
+class CinemaHall:
+    def __init__(self,
+                 number: int):
+        self.number = number
+
+    def movie_session(self,
+                      movie_name: str,
+                      customers: Customer,
+                      cleaning_staff: Cleaner):
+        print(f"\"{movie_name}\" started in hall number {self.number}.")
```
MasterpieceElbow, line 14: Use `f’ “Some string” ‘` if you need to use `”` inside string instead of `f” \” Some string \” “`

MasterpieceElbow (changes requested):

app/main.py
```diff
@@ -1,6 +1,20 @@
...
+        customers_object.append(customer_obj)
+
+    for cust_obj in customers_object:
+        CinemaBar.sell_product(cust_obj, cust_obj.food)
+
+    occupiedCleaner = Cleaner(cleaner)
+    ch = CinemaHall(hall_number)
+
+    # ask mentor why "self" here demands object "ch"
+    CinemaHall.movie_session(ch, movie, customers_object, occupiedCleaner)
```
Chardona, line 20: why it demands 4  parametres here, if CinemaHall.movie_session has 3 arguments?  Parametre ch scores as "self"?

MasterpieceElbow: @Chardona You can write either `Class.movie_session(instance)` or `instance.movie_session()`. Both variants are equivalent. When you write `instance.movie_session()`, instance implicitly passed as self.

MasterpieceElbow (approved):

PR #56, opened 2022-05-23
Solution

Evgeniy994:


app/people/customer.py
```diff
@@ -0,0 +1,7 @@
+class Customer:
+    def __init__(self, name, food):
+        self.name = name
+        self.food = food
+
+    def watch_movie(self, movie):
+        print(f"{self.name} is watching \"{movie}\".")
```
rotsen18, line 7: same

rotsen18 (changes requested):
also probably you should use double "" quotes everywhere

app/people/customer.py
rotsen18: Sorry...my comment disapear. I sad that if you want use "" quotes inside f you should use f' '

rotsen18 (approved):
good job

app/people/customer.py
Evgeniy994: I thought my variant is not a bad practice)
ok i made changes

MasterpieceElbow (approved):

PR #66, opened 2022-06-03
v1

SairusQw:


app/main.py
```diff
@@ -1,6 +1,16 @@
...
+from app.people.customer import Customer
+from app.cinema.bar import CinemaBar
+from app.cinema.hall import CinemaHall
 
 
 def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+    cool_bar = CinemaBar()
+    my_hall = CinemaHall(hall_number)
```
IvanRamyk, line 9: Why do you use the prefix `my_`- for the variable names?

app/main.py
```diff
@@ -1,6 +1,16 @@
...
+from app.people.cinema_staff import Cleaner
+from app.people.customer import Customer
+from app.cinema.bar import CinemaBar
+from app.cinema.hall import CinemaHall
 
 
 def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+    cool_bar = CinemaBar()
```
IvanRamyk, line 8: `cool_bar` isn't a descriptive name. Why is it cool?)

IvanRamyk (changes requested):

PR #81, opened 2022-07-06
Pythons

InnaKushnir:


app/main.py
```diff
@@ -1,6 +1,20 @@
...
+from app.cinema.bar import CinemaBar
+from app.cinema.hall import CinemaHall
+from app.people.cinema_staff import Cleaner
+from app.people.customer import Customer
 
 
 def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+    cust = [i["name"] for i in customers]
```
IvanRamyk, line 8: I believe these lists are redundant and just waste space in the memory

app/main.py
```diff
@@ -1,6 +1,20 @@
...
+    prod = [i["food"] for i in customers]
+    cinema_guests = []
+    cleaner = Cleaner(cleaner)
+    hall_number = CinemaHall(hall_number)
+
+    for i in range(len(customers)):
+        customer = Customer(cust[i], prod[i])
```
IvanRamyk, line 15: You can get the name and the product directly from the `customers ` list 
```
customer = Customer(customers[i]["name"], customers[i]["food"])
```

app/cinema/hall.py
```diff
@@ -0,0 +1,14 @@
...
+from app.people.cinema_staff import Cleaner
+
+
+class CinemaHall:
+    def __init__(self, number):
+        self.number = number
+
+    def movie_session(self, movie_name: str, customers: list, cleaner: str):
+        print(f"\"{movie_name}\" started in hall number {self.number}.")
+        for _ in customers:
```
IvanRamyk, line 11: `_` is often used when we don't use the loop variable and want to emphasize that fact.
But when you pass the variable to method/function it's better to choose a descriptive name for it

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

PR #88, opened 2022-07-11
Solution

DHushchyk:


app/main.py
```diff
@@ -1,6 +1,19 @@
...
 
 
 def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+    customers_list = []
+    for person in customers:
+        person = Customer(name=person["name"], food=person["food"])
+        customers_list.append(person)
+        CinemaBar.sell_product(person.food, person)
```
Abnormaltype, line 13: you have already created 4 `CinemaBar`

app/main.py
```diff
@@ -1,6 +1,19 @@
...
+    CinemaHall.movie_session(
+        CinemaHall(hall_number),
```
Abnormaltype, line 15: you should pass only `hall_number`

Abnormaltype (changes requested):

app/main.py
DHushchyk: But I must add hall_number to CinemaHall class, right? How can I do it? The same question for CinemaBar.

Abnormaltype (changes requested):
comment

app/main.py
DHushchyk: I created instance Cleaner too.

Abnormaltype (approved):
well done!

PR #128, opened 2022-09-08
Solution

vlad-milkov:


app/main.py
```diff
@@ -1,6 +1,47 @@
...
+hall_number = 5
+cleaner_name = "Anna"
+movie = "Madagascar"
+
+cinema_visit(
+    customers=customers,
+    hall_number=hall_number,
+    movie_name=movie,
+    cleaning_staff=cleaner_name
+)
```
DmytroSvirsa, line 47: remove unused (but as note: don't forget about `if __name__ == “main”` when doing like this)

DmytroSvirsa (changes requested):

DmytroSvirsa (approved):

PR #134, opened 2022-09-15
Solution

sbublinskyi:


app/cinema/hall.py
```diff
@@ -0,0 +1,22 @@
...
+    def __init__(self, number: int):
+        self.number = number
+
+    def movie_session(
+            self,
+            movie_name: str,
+            customers: list,
+            cleaning_staff: Cleaner
+    ):
+        print(f"\"{movie_name}\" started in hall number {self.number}.")
```
MasterpieceElbow, line 15: If you have to use `"` inside the string - wrap your string with `'` instead of `"`. Avoid using `\"`. Remake everywhere

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):
GJ!

PR #138, opened 2022-09-16
Solution

nailia-mamiedova:


app/cinema/hall.py
```diff
@@ -0,0 +1,10 @@
+class CinemaHall:
+    def __init__(self, number):
+        self.number = number
+
+    def movie_session(self, movie_name, customers, cleaning_stuff):
+        print(f"\"{movie_name}\" started in hall number {self.number}.")
```
Abnormaltype, line 6: use `''` instead of `""`

app/main.py
```diff
@@ -1,6 +1,16 @@
...
-    # write you code here
-    pass
+    customers_list = [
+        Customer(customer["name"], customer["food"])
+        for customer in customers
+    ]
+    number_hall = CinemaHall(hall_number)
+    cleaner_name = Cleaner(cleaner)
+    for customer in customers_list:
+        CinemaBar.sell_product(customer, customer.food)
```
Abnormaltype, line 15: you have already created 4 different `CinemaBar()`

Abnormaltype (changes requested):

app/cinema/hall.py
nailia-mamiedova: you mean f ' ' or  '{ }'

app/main.py
nailia-mamiedova: I can't find where they are already created

PR #141, opened 2022-09-17
Solution

IvanStored:


StasOkhrym (approved):
Looks great

app/cinema/hall.py
```diff
@@ -0,0 +1,14 @@
+from app.people.cinema_staff import Cleaner
+
+
+class CinemaHall:
+    def __init__(self, number: int):
+        self.number = number
+
+    def movie_session(self, movie_name: str,
+                      customers: list, cleaning_staff: Cleaner) -> None:
+        print(f"\"{movie_name}\" started in hall number {self.number}.")
```
TPolina, line 10: Use `f’ “Some string” ‘` if you need to use `”` inside a string instead of `f” \” Some string \” “`

TPolina (approved):
Well done!

PR #147, opened 2022-09-20
Sansai

Gr1zzZzzly:


app/main.py
```diff
@@ -1,6 +1,18 @@
...
 def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+    obj_customers = [
+        Customer(person["name"], person["food"])
+        for person in customers
+    ]
+
+    [CinemaBar.sell_product(customer, customer.food)
+     for customer in obj_customers]
```
shurushku, line 14: Hi, I don't think it's rational to use list comprehension here. A simple "for" also written in two lines, looks clearer and doesn't create a list of None.

app/cinema/hall.py
```diff
@@ -0,0 +1,13 @@
+class CinemaHall():
+    def __init__(self, number) -> None:
+        self.number = number
+
+    def movie_session(self, movie_name, customers, cleaning_staff):
+
+        print(f'"{movie_name}" started in hall number {self.number}.')
+
+        [customer.watch_movie(movie_name) for customer in customers]
```
shurushku, line 9: Same thing here

app/main.py
Gr1zzZzzly: okey, i fixed it, but the comprehension more beautiful for me

app/cinema/bar.py
```diff
@@ -0,0 +1,5 @@
+class CinemaBar():
```
TPolina, line 1: we don't write brackets when the class takes no arguments. Check this in all your classes

app/main.py
```diff
@@ -1,6 +1,20 @@
...
-    pass
+    obj_customers = [
+        Customer(person["name"], person["food"])
+        for person in customers
+    ]
+
+    for customer in obj_customers:
+        CinemaBar.sell_product(customer, customer.food)
+    # [CinemaBar.sell_product(customer, customer.food)
+    #  for customer in obj_customers]
```
TPolina, line 16: delete unnecessary comments

TPolina (changes requested):

app/main.py
TPolina: we use comprehensions to create list, dict or set. In this task you don't need list, you need just run the method for each element

app/cinema/hall.py
```diff
@@ -0,0 +1,14 @@
+class CinemaHall:
+    def __init__(self, number) -> None:
+        self.number = number
+
+    def movie_session(self, movie_name, customers, cleaning_staff):
```
Aleksandro777, line 5: add annotation

Aleksandro777 (changes requested):

app/cinema/bar.py
```diff
@@ -0,0 +1,5 @@
+class CinemaBar:
+
+    @staticmethod
+    def sell_product(customer, product: str) -> None:
```
Aleksandro777, line 4: You need to add annotation for the function parameters as well. Review, pls, other functions also. I mean (product: str, customer: Customer)

app/cinema/hall.py
```diff
@@ -0,0 +1,19 @@
+class CinemaHall:
+    def __init__(self, number) -> None:
```
Aleksandro777, line 2: and here

app/people/cinema_staff.py
```diff
@@ -0,0 +1,7 @@
+class Cleaner:
+
+    def __init__(self, name) -> None:
```
Aleksandro777, line 3: Here too

app/people/customer.py
```diff
@@ -0,0 +1,8 @@
+class Customer:
+
+    def __init__(self, name, food) -> None:
```
Aleksandro777, line 3: Here too

Aleksandro777 (changes requested):

Aleksandro777 (approved):

app/cinema/bar.py
TPolina: yes)

TPolina (approved):
Well done!

PR #150, opened 2022-09-20
Solution

taras-andruschenko:


app/cinema/bar.py
```diff
@@ -0,0 +1,5 @@
+class CinemaBar:
+
+    @staticmethod
+    def sell_product(product, customer):
```
IvanStored, line 4: Add annotations here and everywhere you missed

app/cinema/hall.py
```diff
@@ -0,0 +1,18 @@
...
+class CinemaHall:
+
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(self, movie_name: str, customers, cleaning_staff):
+        print(f'"{movie_name}" started in hall number {self.number}.')
+
+        for customer in customers:
+            Customer.watch_movie(customer, movie_name)
```
IvanStored, line 14: u can use "customer", not class)
and at 18 line too, just "cleaning_staff" :)

IvanStored (changes requested):

app/cinema/bar.py
taras-andruschenko: Thanks for your suggestion

IvanStored (approved):

TPolina (approved):

app/cinema/hall.py
```diff
@@ -0,0 +1,19 @@
...
+
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(self, movie_name: str,
+                      customers, cleaning_staff) -> None:
+        print(f'"{movie_name}" started in hall number {self.number}.')
+
+        for customer in customers:
+            Customer.watch_movie(customer, movie_name)
```
TPolina, line 15: `watch_movie` is not a classmethod. Why do you use class name and not object to call it?

app/cinema/hall.py
```diff
@@ -0,0 +1,19 @@
...
+
+        print(f'"{movie_name}" ended.')
+
+        Cleaner.clean_hall(cleaning_staff, self.number)
```
TPolina, line 19: same here: you call `clean_hall` method in a wrong way

TPolina (changes requested):
Sorry, I approved accidentally. Please, fix mistakes

TPolina (approved):

PR #152, opened 2022-09-20
Solution

luxuriant777:


hasslieb3 (approved):

app/cinema/bar.py
```diff
@@ -0,0 +1,5 @@
+class CinemaBar:
+
+    @staticmethod
+    def sell_product(product, customer):
```
TPolina, line 4: add type annotation everywhere

TPolina (changes requested):

app/cinema/bar.py
luxuriant777: type annotations were added

Nattalli (approved):
Well done!

PR #156, opened 2022-09-21
Solution

rostyak:


rostyak:
I added imports because flake8 failed with annotations like (customer: Customer and cleaning_staff: Cleaner) or (customer: "Customer" and cleaning_staff: "Cleaner") even using (from __future__ import annotations)

PR #167, opened 2022-09-22
Solution

Kitazuka:


app/cinema/hall.py
```diff
@@ -0,0 +1,17 @@
...
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(self,
+                      movie_name: str,
+                      customers: list,
+                      cleaning_staff: Cleaner,) -> None:
+        print(f'"{movie_name}" started in hall number {self.number}.')
+        for customer in customers:
+            Customer.watch_movie(customer, movie_name)
```
HStakhiv, line 15: I find it weird to call `Customer` because we should have it in `customers`. But I am not sure if it is a mistake

HStakhiv (changes requested):
Looks good, but ...

HStakhiv (approved):
Good!

app/main.py
```diff
@@ -1,6 +1,19 @@
...
-    pass
+def cinema_visit(customers: list,
+                 hall_number: int,
+                 cleaner: str,
+                 movie: str) -> None:
+
+    customers_list = []
+    for customer in customers:
+        customers_list.append(Customer(customer["name"], customer["food"]))
+        customer_info = Customer(customer["name"], customer["food"])
```
TPolina, line 15: Why do you create Customer object twice?

TPolina (changes requested):

TPolina (approved):

PR #171, opened 2022-09-22
'Solution'

IvanGLS:


tests/test_bar.py
```diff
@@ -1,4 +1,3 @@
-import pytest
```
IvanRamyk, line 1: do not modify tests

IvanRamyk (changes requested):

tests/test_bar.py
IvanGLS: Sorry. Thats removed from pycharm when I move progect from VS code. Already fixed. Check please

app/main.py
```diff
@@ -1,6 +1,25 @@
...
+
+    list_of_customers = []
+    for customer_checker in customers:
+        list_of_customers.append(
+            Customer(name=customer_checker["name"],
+                     food=customer_checker["food"])
+        )
+
+    for bar_work in list_of_customers:
+        CinemaBar.sell_product(bar_work, bar_work.food)
```
Abnormaltype, line 17: yu have already created `n` different `CinemaBar`

Abnormaltype (changes requested):
add type annotation

app/main.py
Abnormaltype: fix this

Abnormaltype (changes requested):

app/main.py
```diff
@@ -1,6 +1,28 @@
...
+
+    list_of_customers = []
+    for customer_checker in customers:
+        list_of_customers.append(
+            Customer(name=customer_checker["name"],
+                     food=customer_checker["food"])
+        )
+
+    for bar_work in list_of_customers:
+        Seller.sell_product(bar_work, bar_work.food)
```
Abnormaltype, line 20: you have already created `n` different `CinemaBar`

Abnormaltype (changes requested):

app/main.py
IvanGLS: What does n mean? Send me please example

Abnormaltype: n - it's a number, 1, 2, 3, 4 , etc

IvanGLS: I can't figure out where. What should I do?

Abnormaltype: you should create an instance of `CinemaBar` before a loop

Abnormaltype (approved):
Well done!

PR #172, opened 2022-09-22
'Solution'

linaSam:


app/cinema/bar.py
```diff
@@ -0,0 +1,7 @@
+from app.people.customer import Customer
+
+
+class CinemaBar:
+    @staticmethod
+    def sell_product(customer: Customer, product: str):
```
rostyak, line 6: If you use annotations, use them everywhere, and for output as well. For example, def ...(...) -> None: 
Not only in this function, but also in others

app/main.py
```diff
@@ -1,6 +1,20 @@
...
+from app.people.cinema_staff import Cleaner
 
 
 def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+    current_customers = []
+
+    for customer in customers:
+        current_customers.append(Customer(customer["name"], customer["food"]))
```
rostyak, line 11: You can use list comprehension here

rostyak (changes requested):

rostyak (approved):
Looks good!

Y-Havryliv (approved):
GJ!

PR #173, opened 2022-09-22
'Solution'

Ilyakson:


app/cinema/bar.py
```diff
@@ -0,0 +1,7 @@
+from app.people.customer import Customer
+
+
+class CinemaBar:
+    @staticmethod
+    def sell_product(product: str, customer: Customer):
```
rostyak, line 6: If you use annotations, use them everywhere, and for output as well. For example, def ...(...) -> None:
Not only in this function, but also in others

app/main.py
```diff
@@ -1,6 +1,17 @@
...
+from app.people.customer import Customer
 
 
 def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+    visitors = []
+
+    for customer in customers:
+        visitors.append(Customer(customer["name"], customer["food"]))
```
rostyak, line 11: You can use list comprehension here

app/main.py
```diff
@@ -1,6 +1,17 @@
...
+
+    cinema_bar = CinemaBar()
+    cleaner = Cleaner(cleaner)
```
rostyak, line 14: make variable "hall = CinemaHall(hall_number)", like variables above, and use it

rostyak (changes requested):

rostyak (approved):
Good job!

PR #183, opened 2022-09-22
Solution

FranchukS:


app/cinema/hall.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class CinemaHall:
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(self,
+                      movie_name: str,
+                      customers: list[Customer],
+                      cleaning_staff: Cleaner) -> None:
```
sbublinskyi, line 12: if you transfer args to the next line, its better to do transfer every arg like this:
```
def movie_session(
                      self,
                      movie_name: str,
                      customers: list[Customer],
                      cleaning_staff: Cleaner
       ) -> None:
```

sbublinskyi (approved):
nice

TPolina (approved):
Well done!

PR #189, opened 2022-09-23
Develop

Menselface:


app/cinema/bar.py
```diff
@@ -0,0 +1,9 @@
+from app.people.customer import Customer
+
+
+class CinemaBar:
+    @staticmethod
+    def sell_product(customer: Customer, product: str):
```
mltdll, line 6: I'm pretty sure that type annotation is required, so adding `-> None` would be nice.

app/people/customer.py
```diff
@@ -0,0 +1,7 @@
+class Customer:
+    def __init__(self, name: str, food: str) -> None:
+        self.name = name
+        self.food = food
+
+    def watch_movie(self, movie) -> None:
```
mltdll, line 6: `movie` argument is also missing type annotation.

mltdll (changes requested):

mltdll (approved):

app/cinema/bar.py
```diff
@@ -0,0 +1,9 @@
+from app.people.customer import Customer
+
+
+class CinemaBar:
+    @staticmethod
+    def sell_product(customer: Customer, product: str) -> None:
+        product = product
+        customer = customer
```
Nattalli, line 8: What are these two lines for?

Nattalli: I think it is unnecessary

Nattalli (changes requested):

Nattalli (approved):
Well done!

PR #192, opened 2022-09-26
Solution

mariia-movchan:


Mowka7 (changes requested):
Do you need some help? You have to create 4 more files with names "bar" in package named "cinema"(you also create this package) in "app" folder. and other 3 files like this but another names and packages. And then write classes in this files. So you have to push all 5 files.

Mowka7 (approved):
Well done!

mariia-movchan:
Thanks, but I understood :)
Yesterday I fastly push and forgot to make "git add ." and, in conclusion, I pushed only one file )

Nattalli (approved):
Very good job!

PR #230, opened 2022-11-02
Solution

Foksanv:


myroslavatrschk (approved):

app/cinema/hall.py
```diff
@@ -0,0 +1,21 @@
+from __future__ import annotations
```
wQuelS, line 1: Do you use it? It would be better, I guess to update python to 3.10 or even 3.11, since those versions are stable

app/main.py
```diff
@@ -1,6 +1,21 @@
...
+        customers: list,
+        hall_number: int,
+        cleaner: str,
+        movie: str) -> None:
+    custom_instances = [
+        Customer(customer["name"], customer["food"])
+        for customer in customers
+    ]
+    cinema_hall = CinemaHall(hall_number)
+    cinema_bar = CinemaBar()
```
wQuelS, line 17: It is not really needed, since CinemaBar's `sell_product` is a staticmethod and it is okay to use such methods from class name without creating an instance

wQuelS (changes requested):

wQuelS (approved):
Very good

PR #243, opened 2022-11-12
Develop

MykytaMoshchenko:


Abnormaltype (changes requested):
use '' instead of "" if you need to use "" inside(for name of movie)

MykytaMoshchenko:
> use '' instead of "" if you need to use "" inside(for name of movie)

But I don't need to use "" inside (for name of the movie), for the movie it should be "name of movie".

f' \"{name of movie}\" ' - it is not correct? 

image

Abnormaltype (changes requested):
oh, you can remove \

MykytaMoshchenko:
> oh, you can remove \

Understood, thank you!

app/people/customer.py
```diff
@@ -0,0 +1,10 @@
+class Customer:
+    def __init__(self, name: str, food: str) -> None:
+        self.name = name
+        self.food = food
+
+    def __repr__(self) -> str:
+        return f"{self.name}"
+
+    def watch_movie(self, movie: str) -> None:
+        print(f'{self.name} is watching \"{movie}\".')
```
Abnormaltype, line 10: and here

Abnormaltype (changes requested):

PR #245, opened 2022-11-14
Solution

Vladyslav7125:


app/main.py
```diff
@@ -1,6 +1,22 @@
...
+        movie: str
+) -> None:
+
+    visitors = []
+    hall = CinemaHall(hall_number)
+    cleaner = Cleaner(cleaner)
+    for customer in customers:
+        visitor = Customer(customer["name"], customer["food"])
+        visitors.append(visitor)
+        CinemaBar.sell_product(visitor, visitor.food)
```
Abnormaltype, line 20: yu should create a bar before loop

Abnormaltype (changes requested):

app/main.py
Vladyslav7125: fixed

Abnormaltype (approved):
Well done!

PR #246, opened 2022-11-14
Solution

aarrtemm:


app/cinema/bar.py
```diff
@@ -0,0 +1,11 @@
+from app.people.customer import Customer
+
+
+class CinemaBar:
+
+    def __init__(self) -> None:
+        pass
```
Abnormaltype, line 7: delete it if you don't need it

app/main.py
```diff
@@ -1,6 +1,23 @@
...
+                 hall_number: int,
+                 cleaner: str,
+                 movie: str
+                 ) -> None:
+    list_of_customers = [Customer(name_n_food["name"],
+                                  name_n_food["food"])
+                         for name_n_food in customers]
+
+    for customer in list_of_customers:
+        CinemaBar.sell_product(customer.food, customer)
```
Abnormaltype, line 17: you should create only one `CinemaBar`

Abnormaltype (changes requested):

PR #256, opened 2022-11-27
Cinema

VItalik1292:


app/main.py
```diff
@@ -1,6 +1,26 @@
...
 
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(
+        customers: list,
+        hall_number: int,
+        cleaner: str,
+        movie: str,
+) -> None:
```
Serhii-Leonenko, line 12: Add Customer to your list annotation

```
def cinema_visit(
        customers: list[Customer],
        hall_number: int,
        cleaner: str,
        movie: str,
) -> None:
```

Serhii-Leonenko: The same in other similar cases

app/cinema/hall.py
```diff
@@ -0,0 +1,18 @@
...
+        self.number = number
+
+    def movie_session(
+            self,
+            movie_name: str,
+            customers: list,
+            cleaning_staff: Cleaner,
+    ) -> None:
+        print(f'"{movie_name}" started in hall number {self.number}.')
+        for index in customers:
```
Serhii-Leonenko, line 15: `index` is not a very good name, it's good if you use indexes of range, in this case better to use something like this `for customer in customers`

Serhii-Leonenko: In other similar cases too

Serhii-Leonenko (changes requested):

PR #262, opened 2022-12-03
Solution

andrii-skryl:
It seems that somehow I have changed README.md file. I do not undersatand how I did this and if something need to be done with it.


Serhii-Leonenko (approved):
It's not a big problem with README, but don't do it in the future)

PR #272, opened 2022-12-10
Solution

derkach55:


app/cinema/bar.py
```diff
@@ -0,0 +1,4 @@
+class CinemaBar:
+    @staticmethod
+    def sell_product(customer: object, product: str) -> None:
```
katevasilenko, line 3: It’s picky but could you provide more clear type annotate all over the project, like `customer: Customer`

app/cinema/hall.py
```diff
@@ -0,0 +1,12 @@
+class CinemaHall:
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(
+            self, movie_name: str, customers: list,
+            cleaning_staff: object) -> None:
```
katevasilenko, line 7: And here

katevasilenko (approved):
Good solution but it would be great if you fix type annotation

PR #281, opened 2022-12-21
Restored missing files

olenaliuby:


app/main.py
```diff
@@ -1,6 +1,30 @@
...
+    cleaner_instance = Cleaner(cleaner)
+    hall = CinemaHall(hall_number)
+
+    for customer in customers:
+        # making Customers instances
+        customer = Customer(name=customer["name"], food=customer["food"])
+        customers_list.append(customer)
+
+        # selling food
+        CinemaBar.sell_product(customer=customer.name, product=customer.food)
```
Abnormaltype, line 24: you should create only one `CinemaBar`

tests/test_bar.py
```diff
@@ -15,7 +15,7 @@ def test_cinema_bar_sell_product():
     f = io.StringIO()
 
     with redirect_stdout(f):
-        cb.sell_product(customer=customer, product=customer.food)
+        cb.sell_product(customer=customer.name, product=customer.food)
```
Abnormaltype, line 18: don't change the tests

Abnormaltype (changes requested):

app/main.py
```diff
@@ -1,6 +1,33 @@
...
+        hall_number: int,
+        cleaner: str,
+        movie: str
+) -> None:
+    customers_instances_list = []
+    cleaner_instance = Cleaner(cleaner)
+    cinema_bar_instances = CinemaBar()
+    hall = CinemaHall(hall_number)
+
+    for person in customers:
```
katevasilenko, line 18: It’s okay to do this way, but also you can make list comprehension right on a 13 line

app/main.py
```diff
@@ -1,6 +1,33 @@
...
+        # making Customers instances
```
katevasilenko, line 19: It’s picky, but I would prefer to remove comments

katevasilenko (approved):
Good job. It would be great if you delete comments

app/main.py
olenaliuby: Thank you! It's really looking better now

Abnormaltype (approved):
Good job!

PR #295, opened 2023-01-05
'Solution'

StinGER-ua:


app/cinema/hall.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class CinemaHall:
+
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(self, movie_name: str,
+                      customers: list,
+                      cleaning_staff: Cleaner) -> None:
```
olena-hul, line 11: Move each argument to a separate line or place them together in 1 line. For this case - first option is better because line will be too long if all arguments are placed together

app/main.py
```diff
@@ -1,6 +1,23 @@
...
+                 movie: str) -> None:
+
+    customers_list = []
+
+    cinema_bar = CinemaBar()
+    cinema_hall = CinemaHall(hall_number)
+    cleaning_staff = Cleaner(cleaner)
+
+    for customer in customers:
+        customer_add = Customer(customer["name"], customer["food"])
```
olena-hul, line 19: Name it `new_customer` instead of `customer_add`, sounds more natural

olena-hul (changes requested):

olena-hul (approved):

PR #317, opened 2023-01-25
Solution

anastasiiashchoholieva:


YuriiKindrat (changes requested):
all tests should pass, there is no needed files in ur PR

app/cinema/hall.py
```diff
@@ -0,0 +1,14 @@
...
+        self.number = number
+
+    def movie_session(
+        self,
+        movie_name: str,
+        customers: list,
+        cleaning_staff: str
+    ) -> str:
+        print(f'"{movie_name}" started in hall number {self.number}.')
+        [customer.watch_movie(movie_name) for customer in customers]
```
ArsenPidhoretskyi, line 12: In this way better will be
`list(map(lambda customer: customer.watch_movie(movie_name), customers)) `

app/main.py
```diff
@@ -1,6 +1,21 @@
...
-    pass
+def cinema_visit(
+    customers: list,
+    hall_number: int,
+    cleaner: str,
+    movie: str
+) -> str:
+    customer_list = []
+    for customer in customers:
+        customer_list.append(Customer(customer["name"], customer["food"]))
```
ArsenPidhoretskyi, line 15: Use list comprehensions, also customer_list is not good name
Better customers, but this name already used, than customers_objects

app/main.py
```diff
@@ -1,6 +1,21 @@
...
+    for people in customer_list:
```
ArsenPidhoretskyi, line 16: people -> customer

app/main.py
```diff
@@ -1,6 +1,21 @@
...
+        CinemaBar.sell_product(people, people.food)
+
+    cinema_order = CinemaHall(hall_number)
```
ArsenPidhoretskyi, line 19: cinema_order -> cinema_hall or hall

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,6 +1,24 @@
...
+    movie: str
+) -> str:
+    customers_objects = [
+        Customer(customer["name"], customer["food"])
+        for customer in customers
+    ]
+    [
+        CinemaBar.sell_product(customer, customer.food)
+        for customer in customers_objects
+    ]
```
ArsenPidhoretskyi, line 20: In the future use list(map()) as I say before

ArsenPidhoretskyi (approved):
Good!

PR #321, opened 2023-01-27
Solution

Zoriana-Dvulit:


app/cinema/hall.py
```diff
@@ -0,0 +1,14 @@
+from app.people import Cleaner
+
+
+class CinemaHall:
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(self, movie_name: str, customers: list,
+                      cleaning_staff: Cleaner) -> None:
```
Serhii-Leonenko, line 9: Fix indentation like this, it has better readability

```suggestion
    def movie_session(
    self,
     movie_name: str,
      customers: list,
      cleaning_staff: Cleaner
      ) -> None:

```

Serhii-Leonenko: Also in all other similar cases

Serhii-Leonenko (changes requested):

Serhii-Leonenko (approved):
Good job!

PR #335, opened 2023-02-06
Solution

manjustice:


AndereLion (approved):
All is fine !!!

app/main.py
```diff
@@ -1,6 +1,26 @@
...
+    customers = [
+        Customer(name=customer["name"],
+                 food=customer["food"])
+        for customer in customers
+    ]
+    cinema_hall = CinemaHall(hall_number)
+    cleaner = Cleaner(cleaner)
+
+    for customer in customers:
+        CinemaBar().sell_product(product=customer.food, customer=customer)
```
wQuelS, line 22: create an instance of the `CinemaBar` and use method not from class name, just like u've done with `CinemaHall` and `Cleaner`

wQuelS (changes requested):

wQuelS (changes requested):

wQuelS (approved):
Good job

PR #340, opened 2023-02-07
Solution

NEkropo1:


ZabFTFT (approved):
Nothing to edit.Good job ! I also liked your solution with list comprehension at main file.

app/main.py
```diff
@@ -1,6 +1,21 @@
...
+from app.people.customer import Customer
 
 
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(
+    customers: list[dict], hall_number: int, cleaner: str, movie: str
+) -> None:
+    customers_l = [
```
Y-Havryliv, line 10: your solution is good, but I'm strongly against variable names like this `customers_l`. And yeah, I understand that this is a list, but it's better to name it a list, not l

Y-Havryliv (changes requested):

app/main.py
NEkropo1: I totally agree, it was just a fast decision, for not making name customers, and I forgot to refactor that

Y-Havryliv (approved):
GJ!

PR #342, opened 2023-02-08
solution

usrofgh:


app/main.py
```diff
@@ -1,6 +1,23 @@
...
 
+from app.people.cinema_staff import Cleaner
+from app.people.customer import Customer
 
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+
+def cinema_visit(
+        customers: [dict],
```
kuznetsb, line 9: I don't know is it correct to annotate list[dict] in such way...?

kuznetsb (approved):
Good job, I'm not confident just about customers annotation in cinema_visit()

app/cinema/hall.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+class CinemaHall:
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(
+            self,
+            movie_name: str,
+            customers: [Customer],
```
Y-Havryliv, line 12: it's better to use `list[Customer]`

app/cinema/hall.py
```diff
@@ -0,0 +1,22 @@
...
+            cleaning_staff: Cleaner
+    ) -> None:
+
+        print(f'"{movie_name}" started in hall number {self.number}.')
+
+        for cust in customers:
```
Y-Havryliv, line 18:
```suggestion
        for customer in customers:
```

app/main.py
Y-Havryliv: I agree, would be better

app/main.py
```diff
@@ -1,6 +1,23 @@
...
+        hall_number: int,
+        cleaning_staff: str,
+        movie_name: str,
+) -> None:
+    customers = [Customer(**cust) for cust in customers]
```
Y-Havryliv, line 14: `cust` is not good variable name

app/main.py
```diff
@@ -1,6 +1,23 @@
...
+    cleaner = Cleaner(cleaning_staff)
+
+    cinema_hall = CinemaHall(hall_number)
+    cinema_bar = CinemaBar()
+
+    for cust in customers:
```
Y-Havryliv, line 20: same here

Y-Havryliv (changes requested):

Y-Havryliv (approved):
Nice!

PR #343, opened 2023-02-08
Solution

WDemidenko:


app/cinema/bar.py
```diff
@@ -0,0 +1,9 @@
+from app.people.customer import Customer
+
+
+class CinemaBar:
+    pass
```
SergiiMachulin, line 5: I think pass is redundant here

app/cinema/hall.py
```diff
@@ -0,0 +1,14 @@
+from app.people.cinema_staff import Cleaner
+
+
+class CinemaHall:
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(self, movie_name: str, customers: list,
```
SergiiMachulin, line 8: Suggestion - for customer parameter you could expand annotation type inside list

SergiiMachulin (changes requested):

SergiiMachulin (approved):
Nice!

app/main.py
```diff
@@ -1,6 +1,17 @@
...
 
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(customers: list, hall_number: int,
+                 cleaner: str, movie: str) -> None:
+    people = []
+    for customer in customers:
+        customer = (Customer(customer["name"], customer["food"]))
+        people.append(customer)
```
wQuelS, line 12: can you make it with list comprehension?

app/main.py
```diff
@@ -1,6 +1,17 @@
...
+        CinemaBar()
+        CinemaBar.sell_product(customer, customer.food)
+    cinema = CinemaHall(hall_number)
+    clean = Cleaner(cleaner)
```
wQuelS, line 16: I'd better make `bar` variable as an instance of `CinemaBar` and use `bar.sell_product`. 
In your case you have a bit different code style

wQuelS (changes requested):

wQuelS (approved):
Nice

PR #345, opened 2023-02-08
Solution

kuznetsb:


app/cinema/bar.py
```diff
@@ -0,0 +1,10 @@
+from app.people.customer import Customer
+
+
+class CinemaBar:
+    def __init__(self) -> None:
+        pass
+
```
amber-marichi, line 7: i think according to the task CinemaBar class should only have the static method. please check

amber-marichi (changes requested):

amber-marichi (approved):
nice. approved

wQuelS (approved):
Great

PR #355, opened 2023-02-09
solution

Yulia-Mala:


app/main.py
```diff
@@ -1,6 +1,20 @@
...
+from app.people.cinema_staff import Cleaner
 
+from app.cinema.bar import CinemaBar
+from app.cinema.hall import CinemaHall
 
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+
+def cinema_visit(customers: list,
```
wQuelS, line 8: this type annotation can be more descriptive

app/main.py
```diff
@@ -1,6 +1,20 @@
...
+                 hall_number: int,
+                 cleaner: str,
+                 movie: str) -> None:
+    customer_list = []
```
wQuelS, line 12: I'd better make here list comprehension

app/main.py
```diff
@@ -1,6 +1,20 @@
...
+    hall_cleaner = Cleaner(cleaner)
+    hall = CinemaHall(hall_number)
+    cinema_bar = CinemaBar()
+    for person in customers:
+        client = Customer(person["name"], person["food"])
+        cinema_bar.sell_product(person["food"], client)
```
wQuelS, line 18: and here a loop for selling products

wQuelS (changes requested):
code is nice, just some small hints below

app/main.py
Yulia-Mala: Not very much, I'm afraid, but I've added info that dicts are in list. info from README: "customers - a list of customers, elements are dicts with 'name' and desired 'food' of a customer"

app/main.py
Yulia-Mala: ok

Y-Havryliv (approved):
GJ!

PR #356, opened 2023-02-09
Solution

ada-krav:


Yevheniy-Lototskiy (approved):

app/main.py
```diff
@@ -1,6 +1,19 @@
...
+                 hall_number: int,
+                 cleaner: str,
+                 movie: str) -> None:
+
+    customers_list = [
+        Customer(customer["name"], customer["food"]) for customer in customers
+    ]
+    cinema_hall = CinemaHall(hall_number)
+    cleaner = Cleaner(cleaner)
+    [CinemaBar.sell_product(customer, customer.food)
```
wQuelS, line 17: make an instance of `CinemaBar` and use method not from class name. Just like u've done with `CinemaHall` and `Cleaner`

wQuelS (changes requested):

wQuelS (approved):
Nice work, as always:)

PR #357, opened 2023-02-09
Solution

Anatolii-Poznyak:


oksana0410 (approved):

app/main.py
```diff
@@ -1,6 +1,24 @@
...
+        hall_number: int,
+        cleaner: str,
+        movie: str
+) -> None:
+
+    customers_list = []
+    for i, customer in enumerate(customers):
+        customers_list.append(
+            Customer(customer["name"], customer["food"])
+        )
```
wQuelS, line 18: this can be done with list comprehension

app/main.py
```diff
@@ -1,6 +1,24 @@
...
+        CinemaBar().sell_product(
+            customers_list[i].food, customers_list[i]
+        )
+    CinemaHall(hall_number).movie_session(
+        movie, customers_list, Cleaner(cleaner)
```
wQuelS, line 23: create instances of those classes, 'cause it is better to use method on with them instead of class names

Y-Havryliv (approved):
GJ!

PR #365, opened 2023-02-09
'Solution'

Morty67:


MikhailLyvak (approved):
Не знаю чи це потрібно, але хз чи треба тут перевірки, які ти добавив. Залишим на розсуд менторів)

app/main.py
```diff
@@ -1,6 +1,17 @@
...
-    pass
+def cinema_visit(
+    customers: list, hall_number: int, cleaner: str, movie: str
+) -> None:
+    clean_worker = Cleaner(cleaner)
+    cinema_hall = CinemaHall(hall_number)
+    customers_list = []
+    for customer in customers:
+        customer = Customer(customer["name"], customer["food"])
+        customers_list.append(customer)
```
wQuelS, line 15: this can be done in one line using list comprehension

app/main.py
```diff
@@ -1,6 +1,17 @@
...
+        CinemaBar.sell_product(customer, customer.food)
```
wQuelS, line 16: make an instance of `CinemaBar` and use method not from class name

wQuelS (changes requested):
don't change tests in any way!
I guess u've used `black` on the whole project. It is better to use it on the specific file you need.
Fix everything

PR #367, opened 2023-02-09
Solution

oigod:


NazarHladaniuk (approved):

app/main.py
```diff
@@ -1,6 +1,19 @@
...
+                 hall_number: int,
+                 cleaner: str,
+                 movie: str
+                 ) -> None:
+    customers = [Customer(**customer) for customer in customers]
+    hall_event = CinemaHall(hall_number)
+    hall_cleaner = Cleaner(cleaner)
+
+    for customer in customers:
+        CinemaBar.sell_product(customer, customer.food)
```
wQuelS, line 17: make an instance of `CinemaBar` and call the method using that instance, not a class name

wQuelS (changes requested):

app/main.py
oigod: 3 minutes

wQuelS (approved):
Good job

PR #369, opened 2023-02-09
Solution1

inrock1:


inrock1:
Cheks failed - becouse flake8 want to see single quotes in text print strings and pytest want to see double quotes.
I left the double quotes.

Morty67 (approved):

app/main.py
```diff
@@ -1,6 +1,25 @@
...
+    # make instances of Customers
+    list_customers_ins = [(Customer(customer["name"], customer["food"]))
+                          for customer in customers]
+
+    cin_hall_ins = CinemaHall(hall_number)  # make instance of CinemaHall
+    cin_bar_ins = CinemaBar()               # make instance of CinemaBar
+    cleaner_ins = Cleaner(cleaner)          # make instance of Cleaner
+
+    [cin_bar_ins.sell_product(customer.food, customer)
+     for customer in list_customers_ins]   # cinema bar sell food to customers
```
wQuelS, line 23: remove redundant comments, since code is pretty self-explanatory and there is no need in them
Also give better names for your varibales

app/main.py
```diff
@@ -1,6 +1,25 @@
...
+
+    cin_hall_ins.movie_session(movie, list_customers_ins, cleaner_ins)
```
wQuelS, line 25: not very good name for instance

wQuelS (changes requested):

app/main.py
inrock1: corrected

app/main.py
inrock1: corrected

Y-Havryliv (approved):
Nice!

PR #374, opened 2023-02-09
Develop

MikhailLyvak:


app/cinema/bar.py
```diff
@@ -0,0 +1,7 @@
+from app.people.customer import Customer
+
+
+class CinemaBar:
+    @staticmethod
+    def sell_product(customer: Customer, product: str) -> str:
```
me1nyk, line 6: I think you should change type annotation here, just like in other modules)

me1nyk (changes requested):

me1nyk (approved):
Good job)

app/cinema/hall.py
```diff
@@ -0,0 +1,20 @@
...
+class CinemaHall:
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(
+        self,
+        movie_name: str,
+        customers: list[Customer],
+        cleaning_staff: Cleaner
+    ) -> str:
```
Serhii-Leonenko, line 14: Fix annotation, the method doesn't return anything

app/main.py
```diff
@@ -1,6 +1,23 @@
...
 
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(
+    customers: list[Customer],
+    hall_number: int,
+    cleaning_staff: str,
+    movie_name: str,
+) -> str:
```
Serhii-Leonenko, line 12: The same, function doesn't return anything. Fix it in all similar cases

Serhii-Leonenko (changes requested):

app/cinema/hall.py
YuriiKindrat: u didn't fix this

app/main.py
YuriiKindrat: the same

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #377, opened 2023-02-09
Solution

ramsty:


oksana-feshchenko (approved):

app/cinema/hall.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class CinemaHall:
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(
+            self,
+            movie_name: str,
+            customers: list,
```
Nattalli, line 11: better `list[Customer]`: you should specify as much as possible

app/cinema/hall.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class CinemaHall:
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(
+            self,
+            movie_name: str,
+            customers: Customer,
```
Serhii-Leonenko, line 12: You should use annotation like this.

``
```suggestion
            customers: list[Customer],
```

Serhii-Leonenko (changes requested):

app/cinema/hall.py
ramsty: Done

app/cinema/hall.py
ramsty: Done

PR #378, opened 2023-02-09
'Solution'

ykopatko:


app/main.py
```diff
@@ -1,6 +1,21 @@
...
-    # write you code here
-    pass
+def cinema_visit(
+        customers: list,
+        hall_number: int,
+        cleaner: str,
+        movie: str
+) -> None:
+    clients = []
+    for elem in customers:
```
wQuelS, line 14: elem is not that good variable name

app/main.py
```diff
@@ -1,6 +1,21 @@
...
+        clients.append(Customer(
+            name=elem.get("name"),
+            food=elem.get("food")
+        ))
+    for client in clients:
+        CinemaBar.sell_product(customer=client, product=client.food)
+    CinemaHall(hall_number).movie_session(movie, clients, Cleaner(cleaner))
```
wQuelS, line 21: make instance of this particular class and use `movie_session` method from that method not from class name.
also make instance of the `Cleaner` class

wQuelS (changes requested):

Serhii-Leonenko (approved):
Good job!

PR #402, opened 2023-02-20
'Solution'

svitlana-savikhina:


app/main.py
```diff
@@ -1,6 +1,22 @@
...
-    # write you code here
-    pass
+def cinema_visit(customers: list,
+                 hall_number: int,
+                 cleaner: str,
+                 movie: str
+                 ) -> None:
+    customer_list = []
+    for customer in customers:
+        customer_list.append(Customer(customer["name"], customer["food"]))
```
ArsenPidhoretskyi, line 14: Use list comprehension

app/cinema/hall.py
```diff
@@ -0,0 +1,19 @@
...
+from app.people.cinema_staff import Cleaner
+
+
+class CinemaHall:
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(self,
+                      movie_name: str,
+                      customers: Customer,
```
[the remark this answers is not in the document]
svitlana-savikhina: You are absolutely right

app/main.py
```diff
@@ -1,6 +1,21 @@
...
+from app.cinema.hall import CinemaHall
+from app.people.customer import Customer
+from app.people.cinema_staff import Cleaner
+from app.cinema.bar import CinemaBar
 
 
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(customers: list[Customer],
```
ArsenPidhoretskyi, line 7: As for me here accepts list[dict]

ArsenPidhoretskyi (changes requested):

app/main.py
svitlana-savikhina: You are right, with  `list[Customer]` Pycharm highlighted `Class 'Customer' does not define '__getitem__', so the '[]' operator cannot be used on its instances `but when i fixed it to list[dict] everything was perfect.

app/main.py
```diff
@@ -1,6 +1,21 @@
...
 
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(customers: list[dict],
+                 hall_number: int,
+                 cleaner: str,
+                 movie: str
+                 ) -> None:
+    customer_list = [Customer(customer["name"], customer["food"])
```
Bohdan-Salamakha, line 12: use get() method instead of parenthesis everywhere to avoid KeyError
`customer.get("name")`

Bohdan-Salamakha (changes requested):
v1

olena-hul (approved):

olena-hul:
Good job!

PR #433, opened 2023-03-23
CINEMA_VISIT

volodinsy:


app/cinema/hall.py
```diff
@@ -0,0 +1,17 @@
...
+from app.people.cinema_staff import Cleaner
+from app.people.customer import Customer
+
+
+class CinemaHall:
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(self, movie_name: str,
+                      customers: list | Customer,
```
ArsenPidhoretskyi, line 11: What mean | can you explain?)
It accept list of Customer -> list[Customer]

ArsenPidhoretskyi (changes requested):

PR #435, opened 2023-03-24
Develop

OleksandrYanchuk:


app/main.py
```diff
@@ -1,6 +1,23 @@
...
-    pass
+def cinema_visit(
+        customers: list[dict],
+        hall_number: int,
+        cleaner: str,
+        movie: str
+) -> None:
+    customers_objs = []
+    for customer in customers:
+        people = Customer(customer["name"], customer["food"])
```
olena-hul, line 15: `people` is a bad name which doesn't show the aim of the object it stores. You can call it `new_customer` , for example

Bohdan-Salamakha (approved):

PR #442, opened 2023-03-26
CINEMA VISIT

volodinsy:


katevasilenko (changes requested):
24E42553-9FD7-4DD8-8383-B5087ABBE536
You’ve shouldn’t push your solution if the tests are failed. If you can’t find problem by yourself, you should write to the mentors in Slack. 
Your tests failed because of dots in strings. Please fix it

PR #455, opened 2023-04-04
"Solution"

YegorT:


app/cinema/hall.py
```diff
@@ -0,0 +1,19 @@
...
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(
+            self,
+            movie_name: str,
+            customers: list[Customer],
+            cleaning_staff: Cleaner
+    ) -> None:
+        print(f'\"{movie_name}\" started in hall number {self.number}.')
```
olena-hul, line 15: No need to use `\` before `"` inside strings, because your string is wrapped into single quotes.
When we are using the combination of single and double quotes, we can omit using `\`

app/people/customer.py
```diff
@@ -0,0 +1,7 @@
+class Customer:
+    def __init__(self, name: str, food: str,) -> None:
+        self.name = name
+        self.food = food
+
+    def watch_movie(self, movie: str) -> None:
+        print(f'{self.name} is watching \"{movie}\".')
```
olena-hul, line 7: Same here

olena-hul (changes requested):

olena-hul (changes requested):

app/people/customer.py
Bohdan-Salamakha: not fixed

Bohdan-Salamakha (changes requested):
v2

Bohdan-Salamakha (approved):

PR #461, opened 2023-04-11
Solution

flatchers:


app/cinema/hall.py
```diff
@@ -0,0 +1,16 @@
+from app.people.cinema_staff import Cleaner
+
+
+class CinemaHall:
+    def __init__(self, number: int) -> str:
+        self.number = number
+
+    def movie_session(self,
+                      movie_name: str,
+                      customers: list,
```
ArsenPidhoretskyi, line 10: list of what?

app/main.py
```diff
@@ -1,6 +1,23 @@
...
+        cb.sell_product(customer=Customer(name=customer["name"],
+                                          food=customer["food"]),
+                        product=customer["food"])
+
+    ch = CinemaHall(number=hall_number)
+    ch.movie_session(movie_name=movie,
+                     customers=[Customer(name=customer["name"],
+                                         food=customer["food"])
+                                for customer in customers],
+                     cleaning_staff=Cleaner(cleaner))
```
ArsenPidhoretskyi, line 23: You create two times customer objects. Please do it only once

ArsenPidhoretskyi (changes requested):

app/cinema/hall.py
flatchers: @ArsenPidhoretskyi i think that true will be customers: Customer, but glows yellow as an error.

PR #505, opened 2023-05-19
Solution

yuro740:


app/main.py
```diff
@@ -1,6 +1,30 @@
...
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(
+        customers: list,
+        hall_number: int,
+        cleaner: str,
+        movie: str
+) -> None:
+    bar_as_obj = CinemaBar()
```
ArsenPidhoretskyi, line 13:
```suggestion
    bar_as_obj = CinemaBar()
```
Can you explain why you need _as_obj suffix in this case? I do not see any variable names bar
The same questions to hall_as_obj

app/main.py
```diff
@@ -1,6 +1,30 @@
...
+    hall_as_obj = CinemaHall(hall_number)
+    cleaner_as_obj = Cleaner(cleaner)
```
ArsenPidhoretskyi, line 15:
```suggestion
    cleaner_obj = Cleaner(cleaner)
```

app/main.py
```diff
@@ -1,6 +1,30 @@
...
+    customers_as_objs = []
```
ArsenPidhoretskyi, line 16:
```suggestion
    customers_objs = []
```

ArsenPidhoretskyi (changes requested):

app/main.py
yuro740: I named all variable in one style. If it is problem, i will renamed bar and hall.

Vasyl-Smutok (approved):
Good Job 👍

PR #518, opened 2023-05-31
implemented solution

oleksiikolii:


app/cinema/hall.py
```diff
@@ -0,0 +1,17 @@
...
+
+class CinemaHall:
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(self,
+                      movie_name: str,
+                      customers: list[Customer],
+                      cleaner: Cleaner) -> None:
+        print(f'"{movie_name}\" started in hall number {self.number}.')
```
YuriiKindrat, line 13: why do u need slash here?

YuriiKindrat (approved):

app/cinema/hall.py
oleksiikolii: Ops, forgot to remove, I've used \" first everywhere than flake said remove it, and I forgot in one place

PR #527, opened 2023-05-31
Solution

imelnyk007:


app/cinema/hall.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+class CinemaHall:
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(
+            self,
+            movie_name: str,
+            customers: list,
```
YuriiKindrat, line 11: could u use more descritptive type annotation for customers ?

app/cinema/hall.py
```diff
@@ -0,0 +1,22 @@
...
+    ) -> None:
+
+        print(f'\"{movie_name}\" started in hall number {self.number}.')
+        for customer in customers:
+            print(f'{customer.name} is watching \"{movie_name}\".')
+        print(f'\"{movie_name}\" ended.')
+        print(
+            f"Cleaner {cleaning_staff.name} "
+            f"is cleaning hall number {self.number}."
+        )
```
YuriiKindrat, line 22: could u use cleaner method here?)

app/cinema/hall.py
```diff
@@ -0,0 +1,22 @@
...
+        self.number = number
+
+    def movie_session(
+            self,
+            movie_name: str,
+            customers: list,
+            cleaning_staff: Cleaner
+    ) -> None:
+
+        print(f'\"{movie_name}\" started in hall number {self.number}.')
```
YuriiKindrat, line 15: Use `f’ “Some string” ‘` if you need to use `”` inside a string instead of `f” \” Some string \” “`
fix this everywhere

YuriiKindrat (changes requested):

app/cinema/hall.py
```diff
@@ -0,0 +1,20 @@
...
+            movie_name: str,
+            customers: list[Customer],
+            cleaning_staff: Cleaner
+    ) -> None:
+
+        print(f'"{movie_name}" started in hall number {self.number}.')
+        for customer in customers:
+            print(f'{customer.name} is watching "{movie_name}".')
+        print(f'"{movie_name}" ended.')
+        Cleaner.clean_hall(cleaning_staff, self.number)
```
YuriiKindrat, line 20: u have a Cleaner instance here, could u write like this `cleaning_staff.clean_hall(self.number)`?

app/main.py
```diff
@@ -1,6 +1,22 @@
...
-    pass
+def cinema_visit(
+        customers: list,
+        hall_number: int,
+        cleaner: str,
+        movie: str
+) -> None:
+
+    customers_list = []
+    for cust in customers:
```
YuriiKindrat, line 15: cust is a bad name for variable, it is better to use full names for naming

YuriiKindrat (changes requested):

app/main.py
imelnyk007: I change "cust" to "customer", 
but is it okay that in line 15 "customer" is a dict and in line 16 "customer" is an object of the Customer class?

YuriiKindrat: yep, u are just rewriting this variable )
it ok, if u won't use this var later

PR #531, opened 2023-06-01
Implemented cinema app

Altahoma:


app/main.py
```diff
@@ -1,6 +1,19 @@
...
+from app.people.customer import Customer
+from app.cinema.hall import CinemaHall
 
 
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(
+    customers: list[dict], hall_number: int, cleaner: str, movie: str
+) -> None:
```
YuriiKindrat, line 9: it is better to write one arg per line, for better readability

Altahoma: Is this so bad? I use "Black" formatter for help myself, it provide 3 ways to write func definition: all in one when len <= 79, like this and as you said one arg per line if line 8 still longer then 79.

PR #532, opened 2023-06-01
'Solution'

yakhlamova:


app/cinema/hall.py
```diff
@@ -0,0 +1,17 @@
...
+        self.number = number
+
+    def movie_session(self,
+                      movie_name: str,
+                      customers: list[Customer],
+                      cleaning_staff: Cleaner) -> None:
+        print(f'"{movie_name}" started in hall number {self.number}.')
+        for customer in customers:
+            customer.watch_movie(movie_name)
+        print(f'"{movie_name}\" ended.')
```
YuriiKindrat, line 16: gj, but why do u need `\` here?)

YuriiKindrat (approved):

app/cinema/hall.py
yakhlamova: oh, sorry, I missed it while changing. Firstly I wrote \"{movie_name}\".

PR #537, opened 2023-06-01
Solution for Cinema Visit

OlhaTryhub:


app/cinema/hall.py
```diff
@@ -0,0 +1,21 @@
...
+class CinemaHall:
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(self,
+                      movie_name: str,
+                      customers: list[Customer],
+                      cleaning_staff: Cleaner) -> None:
+
+        print(f""""{movie_name}" started in hall number {self.number}.""")
```
YuriiKindrat, line 14: u can make like this here and in similar cases
`'"movie_name" started in hall .. '`

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #583, opened 2023-07-10
Solution

ivanzh13:


app/cinema/bar.py
```diff
@@ -0,0 +1,10 @@
+from app.people.customer import Customer
+
+
+class CinemaBar:
+    def __init__(self) -> None:
+        ...
+
```
vsmutok, line 7: Why you need it?

vsmutok (changes requested):

app/cinema/bar.py
ivanzh13: wrote first, forgot to delete

ArsenPidhoretskyi (approved):
Well done!

PR #589, opened 2023-07-15
solution

MishkaGAMII:


MishkaGAMII:
I recreated pull request. But in mate is shown my previos pull request. 
Can you fix it ? Or should I add this solution to the old one ?

PR #605, opened 2023-07-30
Solution

AlexPelevin99:


app/main.py
```diff
@@ -1,6 +1,24 @@
...
+    customers: list[dict],
+    hall_number: int,
+    cleaner: str,
+    movie: str
+) -> None:
+    cinema_bar = CinemaBar()
+    cinema_hall = CinemaHall(hall_number)
+    cleaner = Cleaner(cleaner)
+
+    for i in range(len(customers)):
```
pavlejviki, line 17: Use a more descriptive variable name instead of i.

app/main.py
```diff
@@ -1,6 +1,24 @@
...
+        customers[i] = Customer(customers[i]["name"], customers[i]["food"])
```
pavlejviki, line 18: it is generally better to keep the original data intact, so  create a new list for the created Customer objects instead of modifying the original list.

pavlejviki (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #613, opened 2023-08-08
Solution

nyanplague:


RPodolets (approved):

RPodolets (approved):

app/main.py
```diff
@@ -1,6 +1,21 @@
...
+def cinema_visit(customers: list[dict],
+                 hall_number: int,
+                 cleaner: str,
+                 movie: str) -> None:
+    customer_instances = [
+        Customer(customer["name"], customer["food"])
+        for customer in customers
+    ]
+    for customer in customer_instances:
+        cinema_bar = CinemaBar()
```
Y-Havryliv, line 16: There will be a lot of cinema_bars) It'd be better to create it only once and then use it

Y-Havryliv (changes requested):

Y-Havryliv (approved):
nice

PR #617, opened 2023-08-09
solution

innyshka:


app/cinema/hall.py
```diff
@@ -0,0 +1,18 @@
...
+    def movie_session(self,
+                      movie_name: str,
+                      customers: list[Customer],
+                      cleaning_staff: Cleaner) -> None:
+        print(f'"{movie_name}" started in hall number {self.number}.')
+
+        for customer in customers:
+            customer.watch_movie(movie_name)
+        print(f'"{movie_name}" ended.')
+        Cleaner.clean_hall(cleaning_staff, self.number)
```
Y-Havryliv, line 18: you can use `cleaning_staff` instead of `Cleaner` here, since cleanign_staff is Cleaner instance

app/main.py
```diff
@@ -1,6 +1,19 @@
...
+                 hall_number: int,
+                 cleaning_staff: str,
+                 movie_name: str) -> None:
+    customers = [
+        Customer(customer.get("name"), customer.get("food"))
+        for customer in customers
+    ]
+    for customer in customers:
+        CinemaBar.sell_product(customer.food, customer)
+    hall_number = CinemaHall(hall_number)
```
Y-Havryliv, line 18: `hall_number` is a bit misleading because I expect it to be a number from the name, but it's the `CinemaHall` instance

Y-Havryliv (changes requested):

PR #634, opened 2023-08-10
'Solution'

ochernous:


app/cinema/bar.py
```diff
@@ -0,0 +1,7 @@
+from app.people.customer import Customer
+
+
+class CinemaBar:
+    @staticmethod
+    def sell_product(customer: Customer, product: str) -> None:
+        print(f"cinema bar sold {product} to {customer.name}.")
```
IvanRamyk, line 7: you should not modify tests but fix your print message

tests/test_main.py
```diff
@@ -30,8 +30,8 @@
             5,
             "Anna",
             "Madagascar",
-            'Cinema bar sold Coca-cola to Bob.\n'
-            'Cinema bar sold popcorn to Alex.\n'
+            'cinema bar sold Coca-cola to Bob.\n'
```
IvanRamyk, line 33: do NOT modify tests

IvanRamyk (changes requested):

IvanRamyk (approved):

PR #647, opened 2023-08-14
Solution

Ihor-MA:


app/cinema/__init__.py
```diff
@@ -0,0 +1,18 @@
+class CinemaBar:
```
Kroll410, line 1: __init__.py should contain any logic besides overriding importing flow. By the way, do you actually need these classes and functions?

Kroll410 (changes requested):

PR #661, opened 2023-08-25
Develop1

Daniil-Pankieiev:


prochigor (commented):
hall.py cleaning_staff shouldn`t be a list in annotation
hall.py, main.py should change anotations like that
def cinema_visit(
        customers: list,
        hall_number: int,
        cleaner: str,
        movie: str
) -> None:
that easier to read

prochigor (approved):
if you use diferent ' and " you don't need use \

app/cinema/hall.py
```diff
@@ -0,0 +1,17 @@
...
+
+class CinemaHall:
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(self,
+                      movie_name: str,
+                      customers: list[Customer],
+                      cleaning_staff: Cleaner) -> None:
+        print(f'\"{movie_name}\" started in hall number {self.number}.')
```
LiudmylaKulinchenko, line 13: Use `f’ “Some string” ‘` if you need to use `”` inside a string instead of `f” \” Some string \” “`

app/main.py
```diff
@@ -1,6 +1,24 @@
...
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(customers: list,
+                 hall_number: int,
+                 cleaner_name: str,
+                 movie_title: str) -> None:
+    hall = CinemaHall(hall_number)
+    cleaner = Cleaner(cleaner_name)
+    cinema_bar = CinemaBar()
```

app/main.py
```diff
@@ -1,6 +1,24 @@
...
+    visitors = []
```
LiudmylaKulinchenko, line 14: Use list compresion to create this list, it works faster
You can also rename it to `customert`, it could be more descriptive for list of customers

LiudmylaKulinchenko (approved):

PR #665, opened 2023-08-27
Solution

prochigor:


app/cinema/hall.py
```diff
@@ -0,0 +1,21 @@
+from __future__ import annotations
```
LiudmylaKulinchenko, line 1: Why do you need it here?

app/cinema/hall.py
```diff
@@ -0,0 +1,21 @@
...
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(
+            self,
+            movie_name: str,
+            customers: list,
+            cleaning_staff: Cleaner
+    ) -> None:
+        print(f'\"{movie_name}\" started in hall number {self.number}.')
```
LiudmylaKulinchenko, line 15: You don't need to add \ inside f-string if your quotes are different. Fix throughout the code

app/main.py
```diff
@@ -1,6 +1,27 @@
...
+
+    for customer in customers:
+        customer = Customer(customer["name"], customer["food"])
+        cinema_bar.sell_product(customer, customer.food)
+
+    cinema_hall.movie_session(
+        movie,
+        [Customer(name=i["name"],
+                  food=i["food"]
+                  ) for i in customers],
```
LiudmylaKulinchenko, line 26: Fix indentation here, according to the best practices

LiudmylaKulinchenko: braces must be located in one of two options: open and start with the text or have a line break between the text

app/main.py
```diff
@@ -1,6 +1,27 @@
...
-    # write you code here
-    pass
+def cinema_visit(
+        customers: list,
+        hall_number: int,
+        cleaner: str,
+        movie: str
+) -> None:
+
+    cinema_bar = CinemaBar()
```
LiudmylaKulinchenko, line 14: There is no need to create class instance to call static method

LiudmylaKulinchenko (changes requested):

app/cinema/hall.py
prochigor: I changed something here, and it remained, so I don't.

app/cinema/hall.py
prochigor: Oh, I have done the commit, but I changed it and forgot to commit again when I pushed it.

app/main.py
prochigor: I see, I just use Cinemabar when calling sell_product.

app/main.py
```diff
@@ -1,6 +1,25 @@
...
+
+    for customer in customers:
+        customer = Customer(customer["name"], customer["food"])
+        CinemaBar.sell_product(customer, customer.food)
+
+    cinema_hall.movie_session(
+        movie,
+        [Customer(name=i["name"], food=i["food"])
+         for i in customers],
+        cleaning_staff)
```
LiudmylaKulinchenko, line 25:
```suggestion
    cinema_hall.movie_session(
        movie,
        [
            Customer(name=i["name"], food=i["food"])
            for i in customers
        ],
        cleaning_staff
    )
```

Try like that, it will make your code more readable. Also the last bracket should be placed like the first one (starts with a text or have a line break between the text), like I did in the suggestion

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job 🚀

PR #670, opened 2023-09-01
Solution

Artemh59:


app/cinema/bar.py
```diff
@@ -0,0 +1,5 @@
+class CinemaBar:
+
+    @staticmethod
+    def sell_product(product: str, customer: any) -> str:
```
Bohdan-Salamakha, line 4: invalid typehint for customer and return typehint

app/cinema/hall.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class CinemaHall:
+
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(self,
+                      movie_name: str,
+                      customers: list,
```
Bohdan-Salamakha, line 11: add additional typehint for list inner objects like list[YOUR_VALUE_TYPE]
for example: list that contains string elements has typehint list[str]

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+from app.people.cinema_staff import Cleaner
+from app.people.customer import Customer
+from app.cinema.hall import CinemaHall
+from app.cinema.bar import CinemaBar
 
 
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(customers: list,
```
Bohdan-Salamakha, line 7: add additional typehint for list inner objects like list[YOUR_VALUE_TYPE]
for example: list that contains string elements has typehint list[str]

Bohdan-Salamakha (changes requested):

PR #674, opened 2023-09-03
Solution

SerhiiBudko:
Good day! I have a problem with module "main".py! I don't understand why I have an error: "int object is not iterable", because in the text of the task, it was definitely indicated as "list of customers, elements are dicts".
Can you check it and give recommendations how to solve these issues!

### DIFF
--- a/app/cinema/bar.py
+++ b/app/cinema/bar.py
@@ -1,0 +1,7 @@
+from app.people.customer import Customer
+
+
+class CinemaBar:
+    @staticmethod
+    def sell_product(product: str, costumer: "Customer") -> None:
+        print(f"Cinema bar sold {product} to {costumer.name}")
--- a/app/main.py
+++ b/app/main.py
@@ -1,6 +1,29 @@
-# write your imports here
+from app.cinema.bar import CinemaBar
+from app.cinema.hall import CinemaHall
+from app.people.cinema_staff import Cleaner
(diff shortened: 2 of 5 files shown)


app/cinema/bar.py
```diff
@@ -0,0 +1,7 @@
+from app.people.customer import Customer
+
+
+class CinemaBar:
+    @staticmethod
+    def sell_product(product: str, costumer: "Customer") -> None:
```
Dimosphen1, line 6: Use `Customer` class directly in type annotation here without `""`.

Dimosphen1, line 6: Correct your typo in the `costumer` variable naming.

app/main.py
```diff
@@ -1,6 +1,29 @@
...
+        name = costumer["name"]
+        food = costumer["food"]
+
+        customer = Customer(name, food)
+        customers_info.append(costumer)
+
+    for customer in customers_info:
+        cinema_bar.sell_product(customer.name, costumer.food)
+
+    cinema_hall.movie_session(movie_name, costumers_info, cleaner)
```
Dimosphen1, line 29: Resolve the flake8 error and check your spelling for `costumers_info` variable.

Dimosphen1 (changes requested):
`flake8` error resolution was requested.

PR #675, opened 2023-09-05
Solution

dkibalenko:


app/main.py
```diff
@@ -1,6 +1,20 @@
...
 
 
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(customers: list[dict],
+                 hall_number: int,
+                 cleaner: str,
+                 movie: str) -> None:
+    cinema_customers = [Customer(customer["name"], customer["food"])
```
Dimosphen1, line 11: You could initialize the customer instance here as `Customer(**customer)`.

Dimosphen1 (approved):

app/main.py
dkibalenko: Thanks for the hint, dict unpacking looks definitely better here, whereas I used a bit more explicit way.

PR #678, opened 2023-09-07
Implemented Cinema visit.

AndriiHomeniuk1:


app/main.py
```diff
@@ -1,6 +1,26 @@
...
-    # write you code here
-    pass
+def cinema_visit(
+        customers: list,
+        hall_number: int,
+        cleaner: str,
+        movie: str
+) -> None:
+
+    list_customers = [
```
vasylhnatiuk, line 14:
```suggestion
    customer_instances = [
```

vasylhnatiuk: it will be a better name

vasylhnatiuk (changes requested):

app/main.py
AndriiHomeniuk1: Yes, that's better 
fix it

vasylhnatiuk (approved):

PR #682, opened 2023-09-10
'Solution'

TROTPOL:


app/cinema/hall.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class CinemaHall:
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(
+            self,
+            movie_name: str,
+            customers: list,
```
Dimosphen1, line 11: You could be more specific in type annotation here declaring that you expect to obtain a list of `Customer` instances for `customers` here.

app/main.py
```diff
@@ -1,6 +1,24 @@
...
+from app.cinema.hall import CinemaHall
+from app.people.customer import Customer
+from app.people.cinema_staff import Cleaner
 
 
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(
+        customers: list,
```
Dimosphen1, line 8: You could be more specific in type annotation here declaring that you expect to obtain a list of dicts for `customers` here.

Dimosphen1 (approved):
LGTM, well done! Several comments are left for consideration.

PR #684, opened 2023-09-11
solution: cinema visit

sacrosanctry:


app/cinema/hall.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class CinemaHall:
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(self,
+                      movie_name: str,
+                      customers: list[Customer],
+                      cleaner: Cleaner) -> None:
```
Dimosphen1, line 12: You could change your argument's position so that each argument will be present on a new line and will not be placed on the same line with the method name and return type. Sometimes code becomes more readable using this approach.

Dimosphen1 (approved):

PR #685, opened 2023-09-11
Solution

loplicat:


app/main.py
```diff
@@ -1,6 +1,30 @@
...
+        movie: str
+) -> None:
+
+    customers_list = [
+        Customer(name=client["name"], food=client["food"])
+        for client in customers
+    ]
+
+    for client in customers_list:
+        purchase = CinemaBar()
```
Dimosphen1, line 20: You should initialize the `CustomerBar` instance outside of the loop, there is no need to initialize a new instance with each iteration here.

Dimosphen1 (changes requested):
Several changes were requested.

app/main.py
loplicat: Thank you for checking my code. I've already implemented all the changes. Could you please look through it one more time?

PR #686, opened 2023-09-11
Solution

tkachuk2291:


app/cinema/bar.py
```diff
@@ -0,0 +1,7 @@
+from app.people.customer import Customer
+
+
+class CinemaBar:
+    @staticmethod
+    def sell_product(product: int, customer: Customer) -> None:
```
ArsenPidhoretskyi, line 6: product is not int, it is str

app/people/customer.py
```diff
@@ -0,0 +1,7 @@
+class Customer:
+    def __init__(self, name: str, food: str) -> None:
+        self.name = name
+        self.food = food
+
+    def watch_movie(self, movies: str) -> None:
```
ArsenPidhoretskyi, line 6:
```suggestion
    def watch_movie(self, movie: str) -> None:
```

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #687, opened 2023-09-12
'Solution'

nsdNite:


app/main.py
```diff
@@ -1,6 +1,27 @@
...
+from app.cinema.hall import CinemaHall
+from app.people.cinema_staff import Cleaner
+from app.people.customer import Customer
 
 
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(
+        customers: list,
```
Dimosphen1, line 8: You should specify in a type annotation that you expect to obtain a list of dict for `customers`.

app/cinema/hall.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class CinemaHall:
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(
+            self,
+            movie_name: str,
+            customers: list,
```
Dimosphen1, line 11: You should specify in a type annotation that you expect to obtain a list of `Customer` instances for `customers`.

app/main.py
```diff
@@ -1,6 +1,27 @@
...
+        hall_number: int,
+        cleaner: str,
+        movie: str
+) -> None:
+    customers_instances = []
+    for customer_dict in customers:
+        name = customer_dict.get("name")
+        food = customer_dict.get("food")
+        customer = Customer(name, food)
+        customers_instances.append(customer)
```
Dimosphen1, line 18: You could avoid creating three variables `name`, `food`, and `customer` in this loop. Just append the `Customer` instance directly, besides that, you could use ** unpacking during the initialization as well.

app/main.py
```diff
@@ -1,6 +1,27 @@
...
+
+    hall_instance = CinemaHall(hall_number)
```
Dimosphen1, line 20: There is no need to add `_instance` for each class instance, rename these variables, e.g. `cinema_hall`, `cleaner`, ...

Dimosphen1 (changes requested):
Several changes were requested.

app/main.py
nsdNite: implemented pe adding "typing" imports
thanks for the comment

app/cinema/hall.py
nsdNite: implemented pe adding "typing" imports
thanks for the comment

app/main.py
nsdNite: done
thanks for the comment

app/main.py
nsdNite: done
"bar" variable rises flake 8 VNE002 error, renamed to cinema_bar
thanks for the comment

viktoria-rybenchuk (approved):
Nice!

PR #692, opened 2023-09-18
Solution

MatsalakO:


app/cinema/hall.py
```diff
@@ -0,0 +1,18 @@
+from __future__ import annotations
+from app.people.cinema_staff import Cleaner
+
+
+class CinemaHall:
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(self,
```
viktoria-rybenchuk, line 9: It is better to start the opening bracket on the next line and then have self on the following line. Fix in all places

app/main.py
```diff
@@ -1,6 +1,26 @@
...
+                 hall_number: int,
+                 cleaning_staff: str,
+                 movie_name: str
+                 ) -> None:
+    customers_instances = []
+    for customer_dict in customers:
+        name = customer_dict.get("name")
+        food = customer_dict.get("food")
+        customer = Customer(name, food)
+        customers_instances.append(customer)
```
viktoria-rybenchuk, line 17: Use list comprehension here

app/cinema/hall.py
```diff
@@ -0,0 +1,18 @@
...
+                      movie_name: str,
+                      customers: list,
```
viktoria-rybenchuk, line 11: You can be more specific here in type annotations here defining that you expect to obtain a list of Customer instances for customers

viktoria-rybenchuk (changes requested):

app/cinema/bar.py
```diff
@@ -0,0 +1,8 @@
+from __future__ import annotations
+from app.people.customer import Customer
```
viktoria-rybenchuk, line 2: fix imports

app/cinema/hall.py
```diff
@@ -0,0 +1,21 @@
+from __future__ import annotations
+from typing import List
+from app.people.cinema_staff import Cleaner
+from app.people.customer import Customer
```
viktoria-rybenchuk, line 4: fix imports

viktoria-rybenchuk (changes requested):

app/cinema/hall.py
MatsalakO: > fix imports
i don't understand what exact i need to fix

pavlejviki (approved):
Good job!

PR #707, opened 2023-10-03
Solution

Naz-iv:


app/cinema/bar.py
```diff
@@ -0,0 +1,10 @@
+from app.people.customer import Customer
+
+
+class CinemaBar:
+    def __init__(self) -> None:
```
D1st3f, line 5: No need to use empty __init__

D1st3f (approved):

app/main.py
```diff
@@ -1,6 +1,23 @@
...
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(customers: list,
+                 hall_number: int,
+                 cleaner: str,
+                 movie: str) -> None:
+
+    cinema_hall = CinemaHall(hall_number)
+    cinema_bar = CinemaBar()
```

LiudmylaKulinchenko (changes requested):

PR #708, opened 2023-10-03
Solution for py_cinema_visit

wataschi:


app/main.py
```diff
@@ -1,6 +1,23 @@
...
+from app.people.customer import Customer
+from app.people.cinema_staff import Cleaner
+from app.cinema.hall import CinemaHall
+from app.cinema.bar import CinemaBar
 
 
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(
```
pavlejviki, line 7: write each parameter from new line

pavlejviki (changes requested):

app/main.py
wataschi: Done :)

pavlejviki (approved):
Good job!

PR #711, opened 2023-10-04
Solution

mykolamateichuk:


YuliaHladyshkevych (approved):

app/cinema/hall.py
```diff
@@ -0,0 +1,19 @@
...
+
+class CinemaHall:
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(self,
+                      movie_name: str,
+                      customers: list,
+                      cleaning_staff: "Cleaner"
+                      ) -> None:
```
LiudmylaKulinchenko, line 12: fix formatting

app/main.py
```diff
@@ -1,6 +1,24 @@
...
+
+def cinema_visit(customers: list,
+                 hall_number: int,
+                 cleaner: str,
+                 movie: str) -> None:
+    customer_instances = [Customer(customer["name"], customer["food"])
+                          for customer in customers]
+
+    hall_instance = CinemaHall(hall_number)
+    bar_instance = CinemaBar()
```
LiudmylaKulinchenko, line 16: No need to create an instance to call static method

app/main.py
```diff
@@ -1,6 +1,24 @@
...
-    pass
+
+def cinema_visit(customers: list,
+                 hall_number: int,
+                 cleaner: str,
+                 movie: str) -> None:
+    customer_instances = [Customer(customer["name"], customer["food"])
+                          for customer in customers]
+
+    hall_instance = CinemaHall(hall_number)
```
LiudmylaKulinchenko, line 15: u can just name it `hall`. It is OOP (object-oriebted), so hall anyway is some object. If you want to specify, name it cinema_hall
It is similar to name variable `int_instance` and assign it to some integer value
Fix everywhere

LiudmylaKulinchenko (changes requested):

app/cinema/hall.py
LiudmylaKulinchenko, line 12:
```suggestion
    def movie_session(self,
                      movie_name: str,
                      customers: list,
                      cleaning_staff: "Cleaner") -> None:
```

brackets must be located in one of two options: open and start with the text or have a line break between the text

LiudmylaKulinchenko: If this line becomes too long, choose another option

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1,6 +1,24 @@
...
+
+def cinema_visit(customers: list,
+                 hall_number: int,
+                 cleaner: str,
+                 movie: str) -> None:
+    cinema_customers = [Customer(customer["name"], customer["food"])
+                        for customer in customers]
+
+    cinema_hall = CinemaHall(hall_number)
+    cinema_bar = CinemaBar()
```
LiudmylaKulinchenko, line 16: No need to create an instance to call static method

PR #713, opened 2023-10-05
Develop

YuliaHladyshkevych:


petrykivd (changes requested):

app/main.py
```diff
@@ -1,6 +1,21 @@
...
 
 
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(customers: list, hall_number: int,
+                 cleaner: str, movie: str) -> None:
+
+    customers = [Customer(customer["name"], customer["food"])
+                 for customer in customers]
```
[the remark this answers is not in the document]
LiudmylaKulinchenko: @petrykivd You don't need to use it everywhere, it is not a bad practice in all cases
In this case you don't want Customer to have name or food equal to `None`, so it is better to refer by key directly

app/main.py
```diff
@@ -1,6 +1,21 @@
...
+from app.people.cinema_staff import Cleaner
 
 
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(customers: list, hall_number: int,
+                 cleaner: str, movie: str) -> None:
+
+    customers = [Customer(customer.get("name"), customer.get("food"))
```
LiudmylaKulinchenko, line 10: In this case you don't want Customer to have name or food equal to None, so it is better to refer by key directly

app/main.py
```diff
@@ -1,6 +1,21 @@
...
+    cinema_hall = CinemaHall(hall_number)
+    cinema_cleaner = Cleaner(cleaner)
+
+    for customer in customers:
+        CinemaBar.sell_product(customer.food, customer)
+
+    cinema_hall.movie_session(
+        movie_name=movie,
+        customers=customers,
+        cleaning_staff=cinema_cleaner)
```
LiudmylaKulinchenko, line 21: Fix formatting: brackets must be located in one of two options: open and start with the text or have a line break between the text

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Well done! 👏☀🐣

PR #718, opened 2023-10-05
Solution

RabenkoYevhenii:


app/main.py
```diff
@@ -1,6 +1,20 @@
...
+from app.people.cinema_staff import Cleaner
+
+
+def cinema_visit(
+        customers: list,
+        hall_number: int,
+        cleaner: str,
+        movie: str
+) -> None:
+    cinema_bar = CinemaBar()
```

app/main.py
```diff
@@ -1,6 +1,20 @@
...
+    cinema_hall = CinemaHall(hall_number)
+    cleaner = Cleaner(cleaner)
+    customers_value = [Customer(customer.get("name"), customer.get("food"))
```
LiudmylaKulinchenko, line 16: `customers_value` is not clear naming. `customers` is descriptive anough

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Nice! 🐣☀

PR #720, opened 2023-10-05
Solution

Shatrovskyi:


app/cinema/bar.py
```diff
@@ -0,0 +1,9 @@
+from __future__ import annotations
```
LiudmylaKulinchenko, line 1: unneeded import

app/main.py
```diff
@@ -1,6 +1,22 @@
...
-    pass
+def cinema_visit(customers: list,
+                 hall_number: int,
+                 cleaner: str,
+                 movie: str) -> None:
+
+    customers = [Customer(customer_info["name"], customer_info["food"])
+                 for customer_info in customers]
+
+    cinema_bar = CinemaBar()
```
LiudmylaKulinchenko, line 15: No need to create an instance to call static method

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Well done! 🐣

PR #721, opened 2023-10-05
Solution

MiskoRuslan:


DenPrislipskyi (approved):
The best solution!!!

app/main.py
```diff
@@ -1,6 +1,20 @@
...
 
 
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(customers: list,
+                 hall_number: int,
+                 cleaner: str,
+                 movie: str) -> None:
+    cinema_bar = CinemaBar()
```
LiudmylaKulinchenko, line 11: No need to create an instance to call static method

app/main.py
```diff
@@ -1,6 +1,20 @@
...
+    hall = CinemaHall(hall_number)
+    cleaner_obj = Cleaner(cleaner)
+
+    list_of_customers_obj = [Customer(customer["name"], customer["food"])
```
LiudmylaKulinchenko, line 15: Use shorter names. You just repeat yourself
U can use just `customers` here, because  `customer` anyway means you will assign Customer instances to it

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job 👏

PR #723, opened 2023-10-05
Solution

SelFisher1488:


Familenko (approved):

app/main.py
```diff
@@ -1,6 +1,23 @@
...
+        hall_number: int,
+        cleaner: str, movie: str
+) -> None:
+    customers_instances = [
+        Customer(attributes.get("name"), attributes.get("food"))
+        for attributes in customers
+    ]
+    cinema_hall = CinemaHall(hall_number)
+    cinema_cleaner = Cleaner(cleaner)
+    cinema_bar = CinemaBar()
```
LiudmylaKulinchenko, line 18: No need to create an instance to call static method

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1,6 +1,24 @@
...
+        hall_number: int,
+        cleaner: str, movie: str
+) -> None:
+    customers_instances = [
+        Customer(attributes.get("name"), attributes.get("food"))
+        for attributes in customers
+    ]
+
+    for customer in customers_instances:
+        CinemaBar().sell_product(customer, customer.food)
```
LiudmylaKulinchenko, line 18: You still create an instance here. You can just call it from the class

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job 👏

PR #725, opened 2023-10-05
Solution

Andrii-Silenko:


app/main.py
```diff
@@ -1,6 +1,20 @@
...
-    # write you code here
-    pass
+def cinema_visit(customers: list,
+                 hall_number: int,
+                 cleaner: str,
+                 movie: str) -> None:
+    cinema_hall = CinemaHall(hall_number)
+    cinema_bar = CinemaBar()
+    cleaner = Cleaner(cleaner)
+    customers_list = [Customer(customer["name"], customer["food"])
```
roffi37, line 14: customer.get("key") better way to get value instead customer["key"]

roffi37 (approved):
Good overall!

app/main.py
```diff
@@ -1,6 +1,20 @@
...
 
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(customers: list,
+                 hall_number: int,
+                 cleaner: str,
+                 movie: str) -> None:
+    cinema_hall = CinemaHall(hall_number)
+    cinema_bar = CinemaBar()
```
LiudmylaKulinchenko, line 12: No need to create an instance to call static method

app/main.py
LiudmylaKulinchenko: @roffi37 There should not be Customer with name or food equal to `None`, so it is not necessary here

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Nice 🎈🌼

PR #726, opened 2023-10-05
Solution

DenPrislipskyi:


RVChornyy (approved):
you might not use variable bar_instances and in string 16 apply just class CinemaBar method sell_product. You need no instance but just a method.

app/main.py
```diff
@@ -1,6 +1,21 @@
...
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(customers: list,
+                 hall_number: int,
+                 cleaner: str,
+                 movie: str) -> None:
+    bar_instances = CinemaBar()
+
+    customers_instances = []
```
LiudmylaKulinchenko, line 13: Remove `instances` part from variables. You just repeat yourself, `customer` anyway means you will assign Customer instances to it

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Well done! 👏☀🐣

PR #745, opened 2023-10-10
Solution pray for Slaffcheg halilua Amen!!!

hbilous:


app/cinema/bar.py
```diff
@@ -0,0 +1,10 @@
+from app.people.customer import Customer
+
+
+class CinemaBar:
+    @staticmethod
+    def sell_product(customer: str, product: str) -> None:
+        if isinstance(customer, Customer):
+            print(f"Cinema bar sold {product} to {customer.name}.")
```
EdAlekseiev, line 8: Are you sure that `customer` can be an instance of `Customer`?

EdAlekseiev (changes requested):

app/cinema/bar.py
hbilous: pycharm
I don't know but it's work

ArsenPidhoretskyi: Due to examples and description it always must be customer fix your code

app/main.py
```diff
@@ -1,6 +1,18 @@
...
 
 
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(customers: list,
+                 hall_number: int,
+                 cleaner: str,
+                 movie: str) -> None:
+    customer_list = []
```
ArsenPidhoretskyi, line 11: rename to customers_instances

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+    for customer in customers:
+        customer_list.append(Customer(customer["name"], customer["food"]))
+    hall = CinemaHall(hall_number)
+    cleaner_name = Cleaner(cleaner)
+    for customer in customers:
+        CinemaBar.sell_product(customer["name"], customer["food"])
```
ArsenPidhoretskyi, line 17: sell_product must accept customer as object, not it name

ArsenPidhoretskyi (changes requested):

hbilous:
Why in bar.py we use unknown product but not food from Customer class?
We don't use food from Customer class anywhere

app/main.py
YukihiroSM: Please, don't use some_name_list naming. Better to do as Arsen said.

YukihiroSM:
> Why in bar.py we use unknown product but not food from Customer class? We don't use food from Customer class anywhere

I think, in this case it's only to have some consistency of the data. Ideally the food should be set to product that the bar sold and all other mechanics, but the main point of this task is not about bar and customer, so it's simplified here.

YukihiroSM (changes requested):

YukihiroSM (changes requested):
Fix flake please :)

YukihiroSM (approved):
good job! Thanks!

PR #746, opened 2023-10-12
Solution

OleksandrMV:


app/main.py
```diff
@@ -1,6 +1,19 @@
...
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(customers: list[dict],
+                 hall_number: int,
+                 cleaner: str,
+                 movie: str
+                 ) -> None:
+    cinema_hall = CinemaHall(hall_number)
+    customer_list = [
```
YukihiroSM, line 13: Don't use someobjects_list naming, please. It's hard to understand what should be here by name only. Better to use something like customer_instances.

YukihiroSM (changes requested):

app/main.py
```diff
@@ -1,6 +1,19 @@
...
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(customers: list[dict],
+                 hall_number: int,
+                 cleaner: str,
+                 movie: str
+                 ) -> None:
+    cinema_hall_instance = CinemaHall(hall_number)
+    customer_list = [
```
ArsenPidhoretskyi, line 13: rename to customers_instances

app/main.py
```diff
@@ -1,6 +1,19 @@
...
 
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(customers: list[dict],
+                 hall_number: int,
+                 cleaner: str,
+                 movie: str
+                 ) -> None:
+    cinema_hall_instance = CinemaHall(hall_number)
```
ArsenPidhoretskyi, line 12:
```suggestion
    cinema_hall = CinemaHall(hall_number)
```
here is redundant because you do not have already variable named cinema_hall

app/people/customer.py
```diff
@@ -0,0 +1,7 @@
+class Customer:
+    def __init__(self, name: str, food: str) -> None:
+        self.name = name
+        self. food = food
```
ArsenPidhoretskyi, line 4:
```suggestion
        self.food = food

```

ArsenPidhoretskyi (changes requested):

YukihiroSM (approved):

PR #748, opened 2023-10-15
Solution

paashkovaaa:


app/main.py
```diff
@@ -1,6 +1,22 @@
...
+def cinema_visit(
+        customers: list,
+        hall_number: int,
+        cleaner: str,
+        movie: str
+) -> None:
+    hall = CinemaHall(hall_number)
+    cleaner = Cleaner(cleaner)
+
+    customers_list = [
```
ArsenPidhoretskyi, line 16:
```suggestion
    customers_instances = [
```

app/main.py
```diff
@@ -1,6 +1,22 @@
...
+        Customer(customer["name"], customer["food"])
+        for customer in customers
+    ]
+    for customer_data in customers_list:
```
ArsenPidhoretskyi, line 20:
```suggestion
    for customer in customers_list:
```

app/main.py
ArsenPidhoretskyi: _list suffix is redundant

ArsenPidhoretskyi (changes requested):

app/main.py
paashkovaaa: Can you please check this task.

PR #755, opened 2023-10-17
Solution

CrazyHilly:


CrazyHilly:
First puh -> all checks have failed -> Q003 Change outer quotes to avoid escaping inner quotes (while no errors from flake8 locally)
Replaced double to single
Second push -> all checks have failed -> Q000 Single quotes found but double quotes preferred (same error locally)

app/cinema/bar.py
```diff
@@ -0,0 +1,8 @@
+from app.people.customer import Customer
+
+
+class CinemaBar:
+    @staticmethod
+    def sell_product(product: str, customer: Customer | str) -> None:
+        name = customer.name if isinstance(customer, Customer) else customer
+        print(f'Cinema bar sold {product} to {name}.')
```
EdAlekseiev, line 8:
```suggestion
        print(f"Cinema bar sold {product} to {name}.")

```

EdAlekseiev (changes requested):
Fix flake8, you must use " instead of ' where this is possible

app/cinema/bar.py
CrazyHilly: In my initial commits I used **\"** in place of inner quotes but flake8 on github didn't like it. So I replaced outer quotes with single ones everywhere for consistency across the program. 
So what's better:
- use **\"** for inner quotes
- use **'** for outer quotes everywhere
- use **'** for outer quotes only in files where double quotes in the text strings appear?

app/cinema/hall.py
```diff
@@ -0,0 +1,28 @@
...
+            movie_name: str,
+            customers: list[customer.Customer] | list[dict],
+            cleaning_staff: cinema_staff.Cleaner
+    ) -> None:
+
+        print(f'"{movie_name}" started in hall number {self.number}.')
+
+        for person in customers:
+            if isinstance(person, dict):
+                name, food = person['name'], person['food']
```
EdAlekseiev, line 20:
```suggestion
                name, food = person["name"], person["food"]

```

EdAlekseiev (changes requested):
Flake 8 issues

app/cinema/bar.py
EdAlekseiev: Pep8 doesn't contain strict rules about quotes, however you have to use the same code-style everywhere. I think, that you must use " where it is possible.

CrazyHilly: That sounds a bit strange, frankly. I assumed that it should be either one or another but consistent at least in the file but better throughout the program. 
Anyway, I think the best way is to use an escape character. I don't understand why local flake8 check was okay with it but flake8 on github raised errors.

app/main.py
```diff
@@ -1,6 +1,22 @@
...
+from app.cinema.hall import CinemaHall
+from app.people.cinema_staff import Cleaner
+from app.people.customer import Customer
 
 
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(
+        customers: list[dict] | list[Customer],
```
EdAlekseiev, line 8: `customers` can be only list[dict]

app/main.py
```diff
@@ -1,6 +1,22 @@
...
+        hall_number: int,
+        cleaning_staff: str,
+        movie_name: str
+) -> None:
+
+    for person in customers:
+        if isinstance(person, dict):
+            food, name = person["food"], person["name"]
+        if isinstance(person, Customer):
+            food, name = person.food, person.name
```
EdAlekseiev, line 18: You don't need this block code, because `customers` is always list[dict]. You must create objects for each Customer and pass them to functions

app/cinema/hall.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+class CinemaHall:
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(
+            self,
+            movie_name: str,
+            customers: list[customer.Customer] | list[dict],
```
EdAlekseiev, line 12: `customers` can be only list[customer.Customer]

app/cinema/hall.py
```diff
@@ -0,0 +1,28 @@
...
+
+        print(f'"{movie_name}" started in hall number {self.number}.')
+
+        for person in customers:
+            if isinstance(person, dict):
+                name, food = person["name"], person["food"]
+            if isinstance(person, customer.Customer):
+                name, food = person.name, person.food
+
+            new_customer = customer.Customer(name, food)
```
EdAlekseiev, line 24: This code is redundant

app/cinema/hall.py
```diff
@@ -0,0 +1,28 @@
...
+            new_customer.watch_movie(movie_name)
```
EdAlekseiev, line 25:
```suggestion
            person.watch_movie(movie_name)

```

app/cinema/bar.py
```diff
@@ -0,0 +1,8 @@
+from app.people.customer import Customer
+
+
+class CinemaBar:
+    @staticmethod
+    def sell_product(product: str, customer: Customer | str) -> None:
```
EdAlekseiev, line 6:
```suggestion
    def sell_product(product: str, customer: Customer) -> None:

```

app/cinema/bar.py
```diff
@@ -0,0 +1,8 @@
+from app.people.customer import Customer
+
+
+class CinemaBar:
+    @staticmethod
+    def sell_product(product: str, customer: Customer | str) -> None:
+        name = customer.name if isinstance(customer, Customer) else customer
```
EdAlekseiev, line 7:
```suggestion
        name = customer.name

```

EdAlekseiev (changes requested):

app/main.py
CrazyHilly: Added all those extra annotations and related code to pass pytest as I believed it was passing incorrect arguments to methods. Turned out I missed one detail from task description and was creating Customer instances within CinemaHall class, which is incorrect by default, instead of within the main function; and was passing incorrect arguments to methods myself :)
It should all be fixed now.

YukihiroSM (approved):
Looks good to me!

PR #801, opened 2023-11-14
Done

Moonralex:


app/people/customer.py
```diff
@@ -0,0 +1,21 @@
+from __future__ import annotations
+
+
+class Customer:
+    def __init__(self, name: str, food: str) -> None:
+        self.name = name
+        self.food = food
+
+    def watch_movie(self, movie: str) -> None:
+
```
viktoria-rybenchuk, line 10: remove a blank line

app/people/customer.py
```diff
@@ -0,0 +1,21 @@
...
+    @classmethod
+    def customer_extractor(cls, customer_list: list) -> list[Customer]:
+        return [
+            cls(
+                name=customer["name"],
+                food=customer["food"]
+            )
+            for customer in customer_list
+        ]
```
viktoria-rybenchuk, line 21: why do you need this function? If this is the only one place where this conversion is needed, and the logic is straightforward, you could consider creating the `Customer` objects inline in the `cinema_visit` function instead of using `customer_extractor`

viktoria-rybenchuk (changes requested):

viktoria-rybenchuk (approved):
Well done!

PR #805, opened 2023-11-18
Implemented "Cinema visit"

380974261314kv:


app/main.py
```diff
@@ -1,6 +1,23 @@
...
+    customers: list[dict],
+    hall_number: int,
+    cleaner_name: str,
+    movie: str
+) -> None:
+    customers = [Customer(
+        customer["name"],
+        customer["food"]
+    )
+        for customer in customers]
```
Dimosphen1, line 17: It would be better to format this list comprehension in another way to avoid unnecessary line breaking:
```suggestion
    customers = [Customer(customer["name"], customer["food"])
                 for customer in customers]
```
Prefer this way in similar situations in the future.

Dimosphen1 (approved):

app/main.py
380974261314kv: Thanks for your advice. I'll try to do as you.

PR #806, opened 2023-11-19
Cinema visit

Ananiev-Vitalii:


app/main.py
```diff
@@ -1,6 +1,25 @@
...
+
+    for customer_info in customers:
+        customer = Customer(
+            name=customer_info["name"],
+            food=customer_info["food"])
+        cinema_bar.sell_product(customer=customer, product=customer.food)
+
+    cinema_hall.movie_session(
+        movie_name=movie,
+        customers=[Customer(**c) for c in customers],
```
ArsenPidhoretskyi, line 24: Why do you create customers two times?

ArsenPidhoretskyi (changes requested):

PR #808, opened 2023-11-20
Cinema visit implemented

Black3942:


app/main.py
```diff
@@ -1,6 +1,21 @@
...
+def cinema_visit(
+        customers: list,
+        hall_number: int,
+        cleaner: str,
+        movie: str
+) -> None:
+    visitors = [
+        Customer(customer["name"], customer["food"]) for customer in customers
+    ]
+    cinema_bar = CinemaBar()
```
fsocie7y, line 16: No need to create an instance to call static method

fsocie7y (changes requested):

app/main.py
Black3942: refactored

PR #822, opened 2023-11-28
Develop

TymurSh:


app/cinema/bar.py
```diff
@@ -0,0 +1,5 @@
+class CinemaBar:
+
+    @staticmethod
+    def sell_product(product: str, customer: str) -> None:
```
NikOneZ1, line 4: "customer" type should be Customer type
```suggestion
    def sell_product(product: str, customer: Customer) -> None:
```

app/cinema/hall.py
```diff
@@ -0,0 +1,18 @@
...
+from app.cinema.cinema_stuff import Cleaner
+
+
+class CinemaHall:
+
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(
+            self, movie_name: str, customer: list, cleaning_stuff: Cleaner
```
NikOneZ1, line 11: Specify the list type
```suggestion
            self, movie_name: str, customer: list[Customer], cleaning_stuff: Cleaner
```

NikOneZ1 (changes requested):
Fix flake8 and create package people as stated in the task

PR #846, opened 2023-12-16
'Solution'

yablonskyiv:


app/cinema/bar.py
```diff
@@ -0,0 +1,10 @@
+class Customer:
+    def __init__(self, name: str, food: str) -> None:
+        self.name = name
+        self.food = food
```
NikOneZ1, line 4: Why do you need this class here?

app/cinema/hall.py
```diff
@@ -0,0 +1,14 @@
+class CinemaHall:
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(self, movie_name: str,
+                      customers: str, cleaning_staff: str) -> None:
```
NikOneZ1, line 6: 'customers' should be list Customer type and cleaning staff should be Cleaner type
```suggestion
    def movie_session(self, movie_name: str,
                      customers: list[Customer], cleaning_staff: Cleaner) -> None:
```

app/main.py
```diff
@@ -1,6 +1,23 @@
...
+from app.cinema.bar import CinemaBar
+from app.cinema.hall import CinemaHall
+from app.people.customer import Customer
+from app.people.cinema_staff import Cleaner
 
 
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(customers: str, hall_number: int,
```
NikOneZ1, line 7: customers should be list type
```suggestion
def cinema_visit(customers: list[dict], hall_number: int,
```

NikOneZ1 (changes requested):

EdAlekseiev (approved):
Good job

PR #859, opened 2023-12-23
Solution

Vasavik-ua:


app/cinema/hall.py
```diff
@@ -0,0 +1,15 @@
+from app.people.cinema_staff import Cleaner
+from app.people.customer import Customer
+
+
+class CinemaHall:
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(self, movie_name: str, customers: Customer,
```
vasylhnatiuk, line 9: customers is LIST with Customer

app/main.py
```diff
@@ -1,6 +1,16 @@
...
+from app.cinema.bar import CinemaBar
+from app.cinema.hall import CinemaHall
+from app.people.cinema_staff import Cleaner
+from app.people.customer import Customer
 
 
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(customers: list, hall_number: int,
```
vasylhnatiuk, line 7:
```suggestion
def cinema_visit(customers: list[dict], hall_number: int,
```

app/cinema/hall.py
```diff
@@ -0,0 +1,15 @@
...
+                      cleaning_staff: Cleaner) -> None:
+        print(f""""{movie_name}" started in hall number {self.number}.""")
```
vasylhnatiuk, line 11: print(f'\"{movie_name}\" started in hall number {self.number}.')

vasylhnatiuk: use like that

vasylhnatiuk (changes requested):

app/cinema/hall.py
Vasavik-ua: This kind of solution was my first idea, but we don't use the single quats.

app/main.py
```diff
@@ -1,6 +1,16 @@
...
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(customers: list[dict], hall_number: int,
+                 cleaner: str, movie: str) -> None:
+    new_customer = []
+    for client in customers:
+        new_customer.append(Customer(client["name"], client["food"]))
+        CinemaBar.sell_product(Customer(client["name"], client["food"]),
+                               client["food"])
```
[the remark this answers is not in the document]
Vasavik-ua: because i don't save the value of Customer inside the variable...

PR #902, opened 2024-01-15
Develop

MaxOperchuk:


tskopenko (approved):
Еhrough sweat and tears.
Well done!!!

app/main.py
```diff
@@ -4,18 +4,20 @@
...
+        cleaner: str,
+        movie: str
+) -> None:
 
     persons = []
+
     for human in spectators:
         name = human["name"]
         food = human["food"]
         person = Customer(name, food)
```
NEkropo1, line 19: no need for such break in such a simple function and pretty readable names, but overall it's good approach. Still, think about yawni

NEkropo1 (approved):
Good job!

PR #919, opened 2024-01-21
Solution

rozovas:


vasylhnatiuk (changes requested):
fix tests

tests/test_bar.py
```diff
@@ -1,4 +1,3 @@
-import pytest
```
vasylhnatiuk, line 1: why did you delete this?

tests/test_cleaner.py
```diff
@@ -1,4 +1,3 @@
-import pytest
```
vasylhnatiuk, line 1: why did you delete this?

app/main.py
```diff
@@ -1,6 +1,22 @@
...
+from app.cinema.bar import CinemaBar
+from app.cinema.hall import CinemaHall
+from app.people.customer import Customer
+from app.people.cinema_staff import Cleaner
 
 
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(customers: list,
```
```suggestion
def cinema_visit(customers: list[dict],
```

vasylhnatiuk (changes requested):

tests/test_cleaner.py
rozovas: I didn't do it intentionally, not sure how it happened, sorry. I checked it in master and it seems pytest is not used in those files accordingly to Pycharm) the tests seem to work the same with or without these imports, but anyway, I put them back. 
pycharm64_j5V3si6ceK

app/main.py
rozovas: Thanks, done

PR #953, opened 2024-02-04
added solution

MAFIOZNIK1999:


app/cinema/hall.py
```diff
@@ -0,0 +1,16 @@
...
+from app.people.customer import Customer
+
+
+class CinemaHall:
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(self, movie_name: str, customers: list[Customer],
+                      cleaning_staff: Cleaner) -> None:
+        print(f'\"{movie_name}\" started in hall number {self.number}.')
```
LiudmylaKulinchenko, line 11: You can skip `\` in your strings, because your outer quotes are singular (not same as inside a string)

app/main.py
```diff
@@ -1,6 +1,20 @@
...
-    # write you code here
-    pass
+def cinema_visit(customers: list, hall_number: int,
+                 cleaner: str, movie: str) -> None:
+    customers = [
+        Customer(person["name"], person["food"])
+        for person in customers
+    ]
+    cinema_hall = CinemaHall(hall_number)
+    cinema_bar = CinemaBar()
```
LiudmylaKulinchenko, line 14: There is no need to create an instance to use a staticmethod

app/cinema/hall.py
LiudmylaKulinchenko: fix here and in similar cases

LiudmylaKulinchenko (changes requested):

PR #959, opened 2024-02-06
Solution

vanko6226:


YukihiroSM (approved):
Everything looks nice! But please, in future take a look into function arguments alignment. They should be in separate lines: 
https://peps.python.org/pep-0008/#indentation

PR #971, opened 2024-02-13
Solution

d-kaliuta:


app/cinema/bar.py
```diff
@@ -0,0 +1,10 @@
+from app.people.customer import Customer
+
+
+class CinemaBar:
+    @staticmethod
+    def sell_product(customer: Customer | str, product: str) -> None:
```
EdAlekseiev, line 6: Customer should be instance of `Customer` always

EdAlekseiev (changes requested):

app/cinema/bar.py
d-kaliuta: I tried to do as you said, but then I can't pass some tests:

pycharm64_BUzixw5uyz

d-kaliuta: UPD: I found and fixed the error.

PR #989, opened 2024-02-22
Solution

DmytroHlazyrin:


app/people/cinema_staff.py
```diff
@@ -0,0 +1,6 @@
+class Cleaner:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def clean_hall(self, hall_number: int) -> str:
```
nicksetrakov, line 5: your function return nothing , so change type annotation

app/people/customer.py
```diff
@@ -0,0 +1,7 @@
+class Customer:
+    def __init__(self, name: str, food: str) -> None:
+        self.name = name
+        self.food = food
+
+    def watch_movie(self, movie: str) -> str:
```
nicksetrakov, line 6: Here too

nicksetrakov (changes requested):

app/main.py
```diff
@@ -1,6 +1,25 @@
...
+                 hall_number: int,
+                 cleaner: str,
+                 movie: str) -> None:
+
+    customers_list = [
+        Customer(customer["name"], customer["food"])
+        for customer in customers
+    ]
+
+    cinema_bar = CinemaBar()
```
YuriiKindrat, line 17: No need to create an instance to call static method

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,6 +1,25 @@
...
+                 hall_number: int,
+                 cleaner: str,
+                 movie: str) -> None:
+
+    customers_list = [
+        Customer(customer["name"], customer["food"])
+        for customer in customers
+    ]
+
+    cinema_bar = CinemaBar
```
YuriiKindrat, line 17: why do u need this varialbe ?
could u just call method on 23 line like this:
`CinemaBar.sell_product(...)`

YuriiKindrat (changes requested):

app/main.py
DmytroHlazyrin: already! I realized as soon as I asked for a pulrequest.

PR #1004, opened 2024-02-27
'Solution'

ArtomUmantsev:


app/cinema/bar.py
```diff
@@ -0,0 +1,4 @@
+class CinemaBar:
+    @staticmethod
+    def sell_product(product: str, customer: str) -> None:
```
EdAlekseiev, line 3: Invalid type hint for `customer`.
```suggestion
    def sell_product(product: str, customer: Customer) -> None:

```

EdAlekseiev (changes requested):

PR #1043, opened 2024-03-18
Solution

panicua:


Bujhm99 (approved):
image

panicua:
> image


I got what you mean, I was doing word to word what were said in a task, tried not to improvize so nothing breaks with tests :)


p.s. at first I didn't understand your message cause you wrote "You can not to create instance" it's better to write smth like this in English "You don't need to create an instance" cause in your case it's easy to get confused with this part "You can not" (you can't)

app/cinema/hall.py
```diff
@@ -0,0 +1,18 @@
...
+from app.people.customer import Customer
+
+
+class CinemaHall:
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(self,
+                      movie_name: str,
+                      customers: list[Customer],
```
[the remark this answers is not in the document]
YuriiKindrat: `list[Customer]` is perfect )

app/main.py
```diff
@@ -1,6 +1,28 @@
...
+                 movie: str) -> None:
+
+    customers_instances = []
+
+    for customer in customers:
+        new_customer = Customer(customer["name"], customer["food"])
+        customers_instances.append(new_customer)
+
+    hall_instance = CinemaHall(hall_number)
+    bar_instance = CinemaBar()
```
YuriiKindrat, line 19: No need to create an instance to call static method

app/main.py
```diff
@@ -1,6 +1,28 @@
...
+def cinema_visit(customers: list,
+                 hall_number: int,
+                 cleaner: str,
+                 movie: str) -> None:
+
+    customers_instances = []
+
+    for customer in customers:
+        new_customer = Customer(customer["name"], customer["food"])
+        customers_instances.append(new_customer)
```
YuriiKindrat, line 16: could u do this in for loop in 22-23 line ?)

YuriiKindrat (changes requested):

panicua (commented):
@YuriiKindrat , I hope I understood you right with ```for``` loop :)

Oleksl888 (approved):
Good job.
Small remark - I see you you used single quotes for strings to avoid escaping symbols in print statements, but in other files you use double quotes. Best practice is to be consistent.

panicua:
> mall remark - I see you you used single quotes for strings to avoid escaping symbols in print st

I tried to do it like this at first ```f"my name is \"John\""``` but flake8 is agruing with that usage in f-string and says use single quote ;o

Oleksl888:
> > mall remark - I see you you used single quotes for strings to avoid escaping symbols in print st
> 
> I tried to do it like this at first `f"my name is \"John\""` but flake8 is agruing with that usage in f-string and says use single quote ;o

Yeah I get it, this is not really a mistake. What I meant is that if you commit to using single quotes in one file  then it would be good practice to use it the same way in the entire project. No need to change anything right now

PR #1045, opened 2024-03-19
Solution

Line181:


app/main.py
```diff
@@ -1,6 +1,21 @@
...
-    pass
+
+def cinema_visit(
+        customers: list[Customer],
+        hall_number: int,
+        cleaner: str,
+        movie: str) -> None:
+    list_of_customers = []
+    for customer in customers:
+        customer["name"] = Customer(customer["name"], customer["food"])
```
pavlopro, line 16: It's not good to modify the entire dictionary. Instead, create a separate variable `customer `and work with that.

pavlopro (changes requested):

app/main.py
```diff
@@ -1,6 +1,21 @@
...
-    pass
+
+def cinema_visit(
+        customers: list[Customer],
+        hall_number: int,
+        cleaner: str,
+        movie: str) -> None:
+    list_of_customers = []
+    for customer in customers:
+        customer_name = Customer(customer["name"], customer["food"])
```
pavlopro, line 16: This variable holds an object of the Customer class, not just the name of the customer. So, it would be great to rename it to just customer.

pavlopro (changes requested):

pavlopro (approved):
Good Job!

PR #1046, opened 2024-03-19
Solution

DOLC3-GABBANA:


app/main.py
```diff
@@ -1,6 +1,23 @@
...
+from app.cinema.hall import CinemaHall
+from app.people.cinema_staff import Cleaner
+from app.people.customer import Customer
 
 
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(
+    customers: list,
```
pavlopro, line 8: Please make enhancement to the customer annotation as a list of customers

pavlopro (changes requested):
Please remove unnecessary files. There should be only the app package.

PR #1062, opened 2024-03-20
Solution

Nyxan:


semermaks (approved):
Gj)

app/main.py
```diff
@@ -1,6 +1,18 @@
...
-    # write you code here
-    pass
+def cinema_visit(customers: list,
+                 hall_number: int,
+                 cleaner: str,
+                 movie: str) -> None:
+
+    for customer_info in customers:
+        customer = Customer(customer_info["name"], customer_info["food"])
+        CinemaBar().sell_product(customer=customer, product=customer.food)
```
YuriiKindrat, line 14: no need to use brackets here 
u could call this method like this `CinemaBar.sell_product()`

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+                 hall_number: int,
+                 cleaner: str,
+                 movie: str) -> None:
+
+    for customer_info in customers:
+        customer = Customer(customer_info["name"], customer_info["food"])
+        CinemaBar.sell_product(customer=customer, product=customer.food)
+    cinema_hall = CinemaHall(hall_number)
+    customer_list = [Customer(customer["name"], customer["food"])
+                     for customer in customers]
```
YuriiKindrat, line 17: why do u need to create new list here?
could u create a list using for loop before ?

YuriiKindrat (changes requested):

app/cinema/hall.py
```diff
@@ -0,0 +1,17 @@
...
+
+class CinemaHall:
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(self,
+                      movie_name: str,
+                      customers: list[Customer],
+                      cleaning_staff: Cleaner) -> None:
+        print(f'\"{movie_name}\" started in hall number {self.number}.')
```
Oleksl888, line 13: You don't need to escape double quotes since they are inside of single quotes. Also mind that it is a good practice to use only single or only double quotes thoughout the code

Oleksl888 (approved):
Good job!

PR #1066, opened 2024-03-21
'Solution'

afaidiuk:


yossosh (approved):
I like it!

app/main.py
```diff
@@ -1,6 +1,25 @@
...
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(
+        customers: list,
+        hall_number: int,
+        cleaner: str,
+        movie: str
+) -> None:
+    cb = CinemaBar()
```
Oleksl888, line 13: No need to create an instance - you can call static method directly on a class

app/main.py
```diff
@@ -1,6 +1,25 @@
...
+    ch = CinemaHall(hall_number)
```
Oleksl888, line 14: `ch` is not a meaningful name for this instance

Oleksl888 (changes requested):

YuriiKindrat (approved):
LGTM

PR #1069, opened 2024-03-21
'Solution'

yossosh:


afaidiuk (approved):
+

app/cinema/hall.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class CinemaHall:
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(self, movie: str,
+                      customers: List[Customer],
+                      cleaning_staff: Cleaner) -> None:
+        print(f'"{movie}\" started in hall number {self.number}.')
```
Oleksl888, line 13: You don't need to escape double quotes since your string is in single quotes. Check below also

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+from app.cinema.hall import CinemaHall
+from app.people.cinema_staff import Cleaner
+from app.people.customer import Customer
 
 
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(customers: list, hall_number: int,
+                 cleaner: Cleaner | str, movie: str) -> None:
```
Oleksl888, line 8: check what does the task description say about type for cleaner in this function

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+    cinema_bar = CinemaBar()
+    cinema_hall = CinemaHall(hall_number)
+    cleaner = Cleaner(cleaner)
+
+    for customer_data in customers:
+        customer = Customer(name=customer_data["name"],
+                            food=customer_data["food"])
+        cinema_bar.sell_product(customer, customer.food)
+    cinema_hall.movie_session(movie,
+                              [Customer(**c) for c in customers], cleaner)
```
Oleksl888, line 18: you are already creating customers on line 14. Can you optimize this so you don't have to iterate twice?

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,6 +1,19 @@
...
+                 cleaner: str, movie: str) -> None:
+    cinema_bar = CinemaBar()
+    cinema_hall = CinemaHall(hall_number)
+    cleaner = Cleaner(cleaner)
+
+    customers_list = [Customer(name=c["name"], food=c["food"])
+                      for c in customers]
+
+    for customer in customers_list:
+        cinema_bar.sell_product(customer, customer.food)
```
Oleksl888, line 17: You still have two iterations which is not optimal. You can make one for loop where you can create customer, add it to the list and `sell_product` to them. Also you don't need to create cinemabar instance since you can call `sell_product` on a class

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,6 +1,20 @@
...
-    pass
+def cinema_visit(customers: list,
+                 hall_number: int,
+                 cleaner: str,
+                 movie: str) -> None:
+    cinema_hall = CinemaHall(hall_number)
+    cleaning_staff = Cleaner(cleaner)
+
+    customers_list = []
+    for cus in customers:
```
Oleksl888, line 15: `cus` is not a valid name. choose something more appropriate like `customer_data`

Oleksl888 (changes requested):

PR #1072, opened 2024-03-21
Solution v1.01

Chub-andrew:


app/main.py
```diff
@@ -1,6 +1,24 @@
...
-    # write you code here
-    pass
+def cinema_visit(
+        customers: list,
+        hall_number: int,
+        cleaning_staff: str,
+        movie_name: str
+) -> None:
+
+    cinema_bar = CinemaBar()
```
YuriiKindrat, line 14: no need to create instance of cinemabar to call static method

YuriiKindrat (changes requested):

PR #1075, opened 2024-03-22
Solution

fishray33:


NazarSheremet (changes requested):
Need to fix flake8 
image

fishray33:
Finaly no errors in flake

fishray33:
wonder why the first flake error was when I used single outer quotes.

YuriiSmolii (approved):
LGTM

PR #1076, opened 2024-03-22
Solution

siaserb:


sergeybrekhov (approved):
seems good

app/main.py
```diff
@@ -1,6 +1,22 @@
...
+def cinema_visit(customers: list,
+                 hall_number: int,
+                 cleaner: str,
+                 movie: str) -> None:
+
+    list_of_customers = [Customer(customer["name"], customer["food"])
+                         for customer in customers]
+
+    [CinemaBar.sell_product(customer, customer.food)
+     for customer in list_of_customers]
```
YuriiKindrat, line 16: could u use only one for loop here ? 
also it is better to use for loop when there is no need to create list (15-16 lines)

YuriiKindrat (changes requested):

YuriiKindrat (approved):
LGTM

PR #1077, opened 2024-03-23
Solved

serhii-chaban:


vsmutok (changes requested):
Tests Failed!

app/cinema/bar.py
```diff
@@ -0,0 +1,8 @@
+from __future__ import annotations
```
vsmutok, line 1: remove redundant import

vsmutok: fix everywhere

vsmutok (changes requested):

vsmutok (approved):
Great!

serhii-chaban:
I think there is no need to use second argument  "product" in function sell_product.
We sent into this function  Customer instance, which contains name of this product.
image

PR #1082, opened 2024-03-25
Solution

VladyslavKsh:


kssshevchenko (approved):

app/main.py
```diff
@@ -1,6 +1,20 @@
...
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(customers: list,
+                 hall_number: int,
+                 cleaner: str,
+                 movie: str) -> None:
+    customers = [Customer(person["name"], person["food"])
+                 for person in customers]
+    cinema_bar = CinemaBar()
```
YuriiKindrat, line 13: no need to create an instance to call static method

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,6 +1,19 @@
...
 
 
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(customers: list,
+                 hall_number: int,
+                 cleaner: str,
+                 movie: str) -> None:
+    customers = [Customer(person["name"], person["food"])
```
Oleksl888, line 11: You could do this in one iteration. In small task like this it may not make a huge difference but if you work with big data sets this may result in extra unnecessary compute.

Oleksl888 (approved):
gj!

PR #1094, opened 2024-03-30
Solution

ForestGanst:


app/main.py
```diff
@@ -1,6 +1,21 @@
...
+from app.people.customer import Customer
 
 
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(customers: list,
+                 hall_number: int,
+                 cleaner: str,
+                 movie: str) -> None:
```
Dimosphen1, line 10: In such cases, it would be preferable to place each argument on a new line so that there would be no arguments on the same line with opening and closing parenthesis for better readability. Besides that, do not forget to indicate a list of what instances you expect to obtain in the type annotation.

Dimosphen1 (approved):
LGTM, one comment is left for consideration

PR #1096, opened 2024-04-01
solution

NAZ-KOM:


app/main.py
```diff
@@ -1,6 +1,28 @@
...
+from app.cinema.hall import CinemaHall
+from app.people.customer import Customer
+from app.people.cinema_staff import Cleaner
 
 
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(
+        customers: list,
```
```suggestion
        customers: list[dict],
```

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1,6 +1,28 @@
...
+) -> None:
+    linked_hall = CinemaHall(hall_number)
+    linked_customers = []
+    linked_cleaner = Cleaner(cleaner)
+    for customer in customers:
+        linked_customers.append(
+            Customer(customer.get("name"),
+                     customer.get("food")))
+    for customer in linked_customers:
+        CinemaBar.sell_product(customer, customer.food)
```
pavlopro, line 21: Call the `sell_product `method right after the customer is created instead of making a double iteration.

pavlopro (changes requested):

app/main.py
NAZ-KOM: First iteration is for initalize the list of Customer instances. So I can cycle through them in second iteration.

NAZ-KOM: @pavlopro

app/cinema/bar.py
```diff
@@ -0,0 +1,8 @@
+from __future__ import annotations
+from app.people.customer import Customer
```
Dimosphen1, line 2: Add a blank line between different groups of imports according to PEP-8

app/main.py
```diff
@@ -1,6 +1,28 @@
...
+        cleaner: str,
+        movie: str
+) -> None:
+    linked_hall = CinemaHall(hall_number)
+    linked_customers = []
+    linked_cleaner = Cleaner(cleaner)
+    for customer in customers:
+        linked_customers.append(
+            Customer(customer.get("name"),
+                     customer.get("food")))
```
Dimosphen1, line 19: In case there are no arguments on the same line with an opening parenthesis, your closing parenthesis should not have any arguments on the same line as well. Please, resolve this issue

app/main.py
Dimosphen1, line 21: There is no need to have two for loops. Initialize customers and sell products to them within one loop

Dimosphen1 (changes requested):

EdAlekseiev (approved):
Good job

PR #1105, opened 2024-04-05
"Solution"

poplyasita:


app/cinema/hall.py
```diff
@@ -0,0 +1,20 @@
+from app.people.cinema_staff import Cleaner
+
```
Dimosphen1, line 2: There is no need to leave a blank line between imports that belong to the same group. Please, delete redundant blank lines in all places

app/main.py
```diff
@@ -1,6 +1,26 @@
...
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+from app.people.cinema_staff import Cleaner
+
+from app.people.customer import Customer
+
+
+def cinema_visit(
+        customers: list,
```
Dimosphen1, line 11: Indicate that you expect to obtain a list of dicts for `customers` in the type annotation

Dimosphen1 (changes requested):

PR #1110, opened 2024-04-09
Solution

IKruglyak:


YuriiSmolii (changes requested):
Remove all unrelated to project files. Folder `vvv` should be removed from git at all.

YuriiSmolii (changes requested):
Remove python caches from the remote repository.

PR #1119, opened 2024-04-19
Solution

Jejemira:


app/people/cinema_staff.py
```diff
@@ -0,0 +1,7 @@
+class Cleaner:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def clean_hall(self, hall_number: int) -> None:
+        print(f"Cleaner {self.name} is cleaning hall number \
+{hall_number}.")
```
YukihiroSM, line 7: it's better not to use `\` as line separator. 
https://peps.python.org/pep-0008/#maximum-line-length
The preferred way of wrapping long lines is by using Python’s implied line continuation inside parentheses, brackets and braces. Long lines can be broken over multiple lines by wrapping expressions in parentheses. These should be used in preference to using a backslash for line continuation.

YukihiroSM (changes requested):
Else looks pretty good to me!

P.S. Also, you can try to use `black` for python files formatting

vasylhnatiuk (approved):

PR #1126, opened 2024-04-25
triukhan solution

triukhan:


app/main.py
```diff
@@ -1,6 +1,22 @@
...
+def cinema_visit(
+        customers: list[Customer],
+        hall_number: int,
+        cleaner: str,
+        movie: str
+) -> None:
+    customer_list = [Customer(customer["name"], customer["food"])
+                     for customer in customers]
+    cinema_hall = CinemaHall(hall_number)
+    cinema_bar = CinemaBar()
```
LiudmylaKulinchenko, line 16: No need to create a class instance to use staticmethod

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job 🌼

PR #1135, opened 2024-05-06
CinemaSolution

YakovK1808:


YuriiSmolii (changes requested):
image
Fix flake error.

YakovK1808:
Should be fixed in Ver3

PR #1154, opened 2024-05-19
Solution

SuskyiVolodymyr:


app/main.py
```diff
@@ -1,6 +1,32 @@
...
+        customers_instances.append(
+            Customer(name=customer["name"],
+                     food=customer["food"])
+        )
+
+    hall_instance = CinemaHall(hall_number)
+    bar_instance = CinemaBar()
+    cleaner_instance = Cleaner(cleaner)
+
+    for customer in customers_instances:
```
DevSpaciX, line 25: Think how you can do it logic in 1 loop which starting on 15-th line .
This will optimize the complexity of your algorithm

DevSpaciX (changes requested):

app/main.py
```diff
@@ -1,6 +1,31 @@
...
+        cleaner: str,
+        movie: str
+) -> None:
+    hall_instance = CinemaHall(hall_number)
+    bar_instance = CinemaBar()
+    cleaner_instance = Cleaner(cleaner)
+    customers_instances = []
+    for customer in customers:
+        customers_instances.append(
+            Customer(name=customer["name"],
```
DevSpaciX, line 19: You can write it into a variable and then add it to the list, and you don't have to take the last element from the list
For exmpl -> 
`human = Customer(client["name"], client["food"])`

DevSpaciX (changes requested):

DevSpaciX (approved):
Nice !

PR #1155, opened 2024-05-20
Solution

necobra:


app/main.py
```diff
@@ -1,6 +1,20 @@
...
+from app.people.cinema_staff import Cleaner
+from app.people.customer import Customer
 
 
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(customers: list[dict[str, str]], hall_number: int,
+                 cleaner: str, movie: str) -> None:
+    customers_inst = [
```
DevSpaciX, line 9: I understand that you want to use list comprehension but you still make an additional loop on line 17, it’s better to do it in one loop

app/main.py
```diff
@@ -1,6 +1,20 @@
...
+        Customer(customer["name"], customer["food"]) for customer in customers
+    ]
+
+    cinema_bar = CinemaBar()
+    cleaner_inst = Cleaner(cleaner)
+    hall = CinemaHall(hall_number)
+
+    for customer in customers_inst:
```
DevSpaciX, line 17: Just iterate here on `customers` and create customers_list here ( by using append )

app/people/customer.py
```diff
@@ -0,0 +1,7 @@
+class Customer:
+    def __init__(self, name: str, food: str) -> None:
+        self.name = name
+        self.food = food
+
+    def watch_movie(self, movie: str) -> None:
+        print(f"""{self.name} is watching "{movie}".""")
```
DevSpaciX, line 7: Why did you decide to use triple quote syntax ?)) Just do it here as you did in other places

app/cinema/hall.py
```diff
@@ -0,0 +1,18 @@
...
+        self.number = number
+
+    def movie_session(self, movie_name: str, customers: list[Customer],
+                      cleaner: Cleaner) -> None:
+        print(f'\"{movie_name}\" started in hall number {self.number}.')
+
+        for customer in customers:
+            customer.watch_movie(movie_name)
+
+        print(f'\"{movie_name}\" ended.')
```
DevSpaciX, line 16: Why did you use slashes here and on 11 line ? Maybe `print(f'"{movie_name}" ended.')` better ?

DevSpaciX (changes requested):
Algorithm complexity : 2 * O(n) instead of simple O(n)
And some little optimisation requested

DevSpaciX (approved):
Good job!

PR #1156, opened 2024-05-21
solution

AndrewwSSS:


DevSpaciX (changes requested):
Seems like you forgot to add your changes to commit ? Cuz I see blank files , also tests have not passed ((

app/cinema/hall.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class CinemaHall:
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(self, movie_name: str,
+                      customers: list,
+                      cleaning_staff: str) -> None:
+        print(f'\"{movie_name}\" started in hall number {self.number}.')
```
DevSpaciX, line 11: Use `f’ “Some string” ‘` if you need to use `”` inside a string instead of `f” \” Some string \” “`

DevSpaciX (changes requested):
Pay attention to the comment and change it in all places )

app/people/customer.py
```diff
@@ -0,0 +1,7 @@
+class Customer:
+    def __init__(self, name: str, food: str) -> None:
+        self.name = name
+        self.food = food
+
+    def watch_movie(self, movie: str) -> None:
+        print(f'{self.name} is watching \"{movie}\".')
```
DevSpaciX, line 7: One more place with this break line : ((
Before push just review yourself !

DevSpaciX (changes requested):
Next time pay more attention to details

DevSpaciX (changes requested):
Fix pls bad formating with breaking lines ( \ )

DevSpaciX (approved):

PR #1175, opened 2024-05-24
solution

NikitaNikolaitsev:


app/cinema/hall.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class CinemaHall:
+    def __init__(self, number: int, ) -> None:
+        self.number = number
+
+    def movie_session(
+            self, movie_name: str,
+            customers: list,
+            cleaning_staff: Any) -> None:
```
DevSpaciX, line 11: `cleaning_staff` is `Any` type ? Maybe it is type of `Cleaner` ?

app/cinema/hall.py
```diff
@@ -0,0 +1,16 @@
+from typing import Any
+
+
+class CinemaHall:
+    def __init__(self, number: int, ) -> None:
+        self.number = number
+
+    def movie_session(
+            self, movie_name: str,
+            customers: list,
```
DevSpaciX, line 10: Also it is better to write a little more precisely -> `list[Customer]`

app/cinema/bar.py
```diff
@@ -0,0 +1,8 @@
+from typing import Any
+
+
+class CinemaBar:
+
+    @staticmethod
+    def sell_product(product: str, customer: Any) -> None:
```
DevSpaciX, line 7: Are you sure that `customer` can be `Any` type ?

DevSpaciX (changes requested):

DevSpaciX (approved):
Nice !

PR #1176, opened 2024-05-24
Solution

Andrii1000:


app/main.py
```diff
@@ -1,6 +1,27 @@
...
 
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(
+        customers: list, hall_number: int,
+        cleaner: str, movie: str
+) -> None:
+
+    instances = [
```
DevSpaciX, line 12: Try to merge the list comprehension with the loop on line 17. I mean to implement this logic in one loop instead of creating two. Keep track of the algorithm's complexity.

DevSpaciX (changes requested):

DevSpaciX (approved):
Great !

PR #1177, opened 2024-05-24
'Solution'

BohdanKuzik:


app/main.py
```diff
@@ -1,6 +1,24 @@
...
 
 
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(
+        customers: list, hall_number: int, cleaner: str, movie_name: str
+) -> None:
+    customer_instances = [
+        Customer(name=c["name"], food=c["food"]) for c in customers
```
DevSpaciX, line 11: Try to merge the list comprehension with the loop on line 17. I mean to implement this logic in one loop instead of creating two. Keep track of the algorithm's complexity.

DevSpaciX (changes requested):

app/cinema/hall.py
```diff
@@ -0,0 +1,15 @@
+from app.people.cinema_staff import Cleaner
+
+
+class CinemaHall:
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(
+            self, movie_name: str, customers: list, cleaning_staff: Cleaner
```
DevSpaciX, line 9: Also it is better to write a little more precisely -> ` customers: list[Customer]`

DevSpaciX (approved):
Good job!

PR #1178, opened 2024-05-24
Done Task CImena app

TkachukTara:


app/main.py
```diff
@@ -1,6 +1,26 @@
...
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(
+        customers: list,
+        hall_number: int,
+        cleaner: str,
+        movie: str) -> None:
+
+    cinema_customers = [Customer(name=customer["name"], food=customer["food"])
```
DevSpaciX, line 13: Try to merge the list comprehension with the loop on line 19. I mean to implement this logic in one loop instead of creating two. Keep track of the algorithm's complexity.

DevSpaciX (changes requested):

app/cinema/hall.py
```diff
@@ -0,0 +1,16 @@
+from app.people.cinema_staff import Cleaner
+
+
+class CinemaHall:
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(self,
+                      movie_name: str,
+                      customers: list,
```
DevSpaciX, line 10: Also it is better to write a little more precisely -> `customers: list[Customer]`

DevSpaciX (approved):

PR #1180, opened 2024-05-24
Solution

b3v3kt0r:


app/main.py
```diff
@@ -1,6 +1,21 @@
...
 
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(customers: list,
+                 hall_number: int,
+                 cleaner: str,
+                 movie: str) -> None:
+
+    customer_instances = [Customer(customer["name"], customer["food"])
```
DevSpaciX, line 12: Try to merge the list comprehension with the loop on line 18. I mean to implement this logic in one loop instead of creating two. Keep track of the algorithm's complexity.

DevSpaciX (changes requested):
Little optimization requested : )

DevSpaciX (approved):
Good job!

PR #1183, opened 2024-05-24
'Solution'

ghost:


app/cinema/hall.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class CinemaHall:
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(self,
+                      movie_name: str,
+                      customers: list,
+                      cleaning_staff: Any) -> None:
```
DevSpaciX, line 11: `cleaning_staff` is Any type ? Maybe it is type of `Cleaner` ?

app/cinema/hall.py
```diff
@@ -0,0 +1,17 @@
+from typing import Any
+
+
+class CinemaHall:
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(self,
+                      movie_name: str,
+                      customers: list,
```
DevSpaciX, line 10: Also it is better to write a little more precisely -> `list[Customer]`

DevSpaciX (changes requested):
More attention to details and will be perfect !

DevSpaciX (approved):
Great !

PR #1184, opened 2024-05-24
'Solution'

Oleh323:


app/cinema/hall.py
```diff
@@ -0,0 +1,16 @@
+from app.people.cinema_staff import Cleaner
+
+
+class CinemaHall:
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(self,
+                      movie_name: str,
+                      customers: list,
```
DevSpaciX, line 10: It is better to write a little more precisely -> list[Customer]

DevSpaciX (approved):
All good ! But I left the little comment : )

PR #1185, opened 2024-05-25
sol1

Unlie9:


app/main.py
```diff
@@ -1,6 +1,22 @@
...
+from app.cinema.hall import CinemaHall
 
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+
+def cinema_visit(customers: list, hall_number: int,
+                 cleaning_staff: str, movie_name: str) -> None:
+
+    customers = [Customer(name=customer["name"], food=customer["food"]
```
DevSpaciX, line 11: Try to merge the list comprehension with the loop on line 18. I mean to implement this logic in one loop instead of creating two. Keep track of the algorithm's complexity.

app/cinema/hall.py
```diff
@@ -0,0 +1,17 @@
+from app.people.cinema_staff import Cleaner
+
+
+class CinemaHall:
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(self, movie_name: str,
+                      customers: list, cleaning_staff: Cleaner) -> None:
```
DevSpaciX, line 9: It is better to write a little more precisely -> `customers: list[Customer]`

tests/test_cleaner.py
```diff
@@ -1,4 +1,3 @@
-import pytest
```
DevSpaciX, line 1: Pls do not touch the tests files )

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #1216, opened 2024-06-18
implement Cinema visit

HalynaRohatska:
created 2 package


oliasymak24 (approved):
Good job!
P.s. Try to remove unnecessary test comments :)
I like using from `__future__ import annotations` ; it makes the code more error-resistant.

PR #1222, opened 2024-06-23
Implement Solution

SkepskyiDanylo:


app/cinema/bar.py
```diff
@@ -0,0 +1,8 @@
+from __future__ import annotations
```
vsmutok, line 1: I think this import redundant here

app/cinema/hall.py
```diff
@@ -0,0 +1,19 @@
+from __future__ import annotations
```
vsmutok, line 1: I think this import redundant here

vsmutok (changes requested):

app/cinema/hall.py
SkepskyiDanylo: i think you are right

vsmutok (approved):
Great!

PR #1226, opened 2024-06-25
implemented small cinema ecosystem with watching films and buying foo…

MasakDirt:
…d dependencies

### DIFF
--- a/app/cinema/bar.py
+++ b/app/cinema/bar.py
@@ -1,0 +1,7 @@
+from app.people.customer import Customer
+
+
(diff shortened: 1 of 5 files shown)


oliasymak24 (approved):
Good job!
P.s. The constructor always returns None by default, so the return type can be omitted.

MasakDirt:
> Good job! P.s. The constructor always returns None by default, so the return type can be omitted.

Thanks,
Yea, but _flake8_ write that I must write

PR #1234, opened 2024-07-02
The function ensures the correct sequence of actions in the cinema fr…

MaksimGolev:
…om selling food to cleaning the hall after the end of the show


NazarSheremet (approved):
Good job!

PR #1240, opened 2024-07-08
Implemented solution

dimak20:


app/main.py
```diff
@@ -1,6 +1,16 @@
...
+from app.people.cinema_staff import Cleaner
 
 
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(customers: list, hall_number: int,
+                 cleaner: str, movie: str) -> None:
+    customers_list = [Customer(customer["name"], customer["food"])
+                      for customer in customers]
```
MNDonut, line 10: it's better to move it on the next line

app/main.py
```diff
@@ -1,6 +1,16 @@
...
+    for customer in customers_list:
+        temp_bar = CinemaBar()
```
MNDonut, line 12: It's better to move it outside of loop because if there are 10 000 customers, it'll create 10 000 bars for each customer just in order to sell product. Thus some memory will be taken by it

MNDonut (changes requested):

app/cinema/bar.py
```diff
@@ -0,0 +1,7 @@
+from app.people.customer import Customer
+
+
+class CinemaBar:
+
+    def sell_product(self, customer: Customer, product: str) -> None:
```
oliasymak24, line 6: Pay attention to the description of this method in the task. Why did you decide to change it?

oliasymak24 (changes requested):
You changed the previous comments well, but it seems that you need to pay attention again to the types of class methods.

app/cinema/bar.py
dimak20: I came to that conclusion while fixing static method in other tasks. I thought that I didn't need to use instance of a class (and then non-static method) when don't refer to instance's attribute.

oliasymak24 (approved):
Good job!

PR #1257, opened 2024-07-18
Solution

DSahalatyi:


app/cinema/bar.py
```diff
@@ -0,0 +1,12 @@
+from __future__ import annotations
+
+from app.people.customer import Customer
+
+
+class CinemaBar:
+    @staticmethod
+    def sell_product(product: str, customer: str | Customer) -> None:
```
LiudmylaKulinchenko, line 8: `class should have only one static method sell_product, that takes product - name of the product that customer wants and customer - Customer instance, that means customer` -- the class should only take Customer instance, according to the requirements

app/main.py
```diff
@@ -1,6 +1,24 @@
...
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(
+        customers: list,
+        hall_number: int,
+        cleaner: str,
+        movie: str
+) -> None:
+    cb = CinemaBar()
```
LiudmylaKulinchenko, line 13: No need to create instance to call static method

app/main.py
```diff
@@ -1,6 +1,24 @@
...
+    ch = CinemaHall(hall_number)
+    staff = Cleaner(cleaner)
```
LiudmylaKulinchenko, line 15: Use descriptive variable names

app/cinema/hall.py
```diff
@@ -0,0 +1,18 @@
...
+            self,
+            movie_name: str,
+            customers: list,
+            cleaning_staff: Cleaner
+    ) -> None:
+        print(f'"{movie_name}" started in hall number {self.number}.')
+        for customer in customers:
+            customer.watch_movie(movie_name)
+        print(f'"{movie_name}" ended.')
+        cleaning_staff.clean_hall(self.number)
```
LiudmylaKulinchenko, line 18: Let's use blank lines to separate logic in this function a little bit, to make it more readable

LiudmylaKulinchenko (changes requested):

MNDonut (approved):

PR #1261, opened 2024-07-21
Resolved Cinema Visit

viktor-77:


app/main.py
```diff
@@ -1,6 +1,22 @@
...
+def cinema_visit(
+    customers: list, hall_number: int,
+    cleaner: str, movie: str
+) -> None:
+    cinema_customers = []
+    for customer in customers:
+        cinema_customers.append(Customer(customer["name"], customer["food"]))
+
+    cinema_hall = CinemaHall(hall_number)
+    cinema_bar = CinemaBar()
```
pavlopro, line 16: No need to have instance of CinemaBar. Since `sell_product `is a static method u can use class `CinemaBar` instead

pavlopro (changes requested):

pavlopro (approved):
Good Job!

PR #1262, opened 2024-07-22
solution

BastovOleksandr:


app/main.py
```diff
@@ -1,6 +1,22 @@
...
+    movie: str
+) -> None:
+    customers = [
+        Customer(customer["name"], customer["food"]) for customer in customers
+    ]
+    hall = CinemaHall(hall_number)
+    cleaner = Cleaner(cleaner)
+
+    for customer in customers:
+        CinemaBar().sell_product(customer, customer.food)
```
LiudmylaKulinchenko, line 20: No need to create class instance to call staticmethod

LiudmylaKulinchenko (changes requested):
Tests didn't pass

MNDonut (approved):

PR #1264, opened 2024-07-22
Develop

Narberal90:


Narberal90:
well, my local flake8 didn't know that if I escape the double inner quotes - the outer should be single -_-

PR #1266, opened 2024-07-23
Solution

Atikiho:


app/cinema/hall.py
```diff
@@ -0,0 +1,21 @@
...
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(self,
+                      movie_name: str,
+                      customers: list,
+                      cleaning_staff: cinema_staff.Cleaner) -> None:
+
+        print(f'"{movie_name}" started in hall number {self.number}.')
+        for someone in customers:
```
AnyoneClown, line 16: Bad variable name. Use the singular when iterating. For example: `for car in cars`

app/cinema/hall.py
```diff
@@ -0,0 +1,21 @@
...
+            one = customer.Customer(someone.name, "")
+            one.watch_movie(movie_name)
```
AnyoneClown, line 18: Why create another one from an already created customer? This is already an instance of the Customer class, so you should use hist methods right away.

app/cinema/hall.py
```diff
@@ -0,0 +1,21 @@
+from __future__ import annotations
+import app.people.customer as customer
+import app.people.cinema_staff as cinema_staff
```
AnyoneClown, line 3: It is better to import a whole class and use it without `as`.

AnyoneClown (changes requested):

app/main.py
```diff
@@ -1,6 +1,16 @@
...
+def cinema_visit(customers: list,
+                 hall_number: int,
+                 cleaner: str,
+                 movie: str) -> None:
+    all_customers = [Customer(someone["name"], someone["food"])
+                     for someone in customers]
+    for someone in all_customers:
+        CinemaBar().sell_product(someone, someone.food)
+    cinema_hall = CinemaHall(hall_number)
+    cinema_hall.movie_session(movie, all_customers, Cleaner(cleaner))
```
AnyoneClown, line 16: You are not declaring a variable with `Cleaner` instance, but at the same time you are doing this for cinema_hall. We can use this instance without a variable

app/cinema/hall.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class CinemaHall:
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(self,
+                      movie_name: str,
+                      customers: list,
+                      cleaning_staff: app.people.cinema_staff.Cleaner) -> None:
```
AnyoneClown, line 13: Pls, start all arguments from a new line if you have many of them(Check all funcs in your code), for example:

```
def movie_session(
    self,
    movie_name: str
) -> None:
```

PR #1267, opened 2024-07-23
Develop

Narberal90:


app/cinema/hall.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class CinemaHall:
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(self,
+                      movie_name: str,
+                      customers: list["Customer"],
+                      cleaning_staff: "Cleaner") -> None:
```
MNDonut, line 12: The best practice is when parameters start from the next line after method name and the closing bracket is on a separate line

app/main.py
```diff
@@ -1,6 +1,26 @@
...
+from app.people.customer import Customer
 
 
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(customers: list,
+                 hall_number: int,
+                 cleaner: str,
+                 movie: str) -> None:
```
MNDonut, line 10: The best practice is when parameters start from the next line after method name and the closing bracket is on a separate line

MNDonut (changes requested):

app/main.py
```diff
@@ -1,6 +1,27 @@
...
+    customers = [
+        Customer(
+            name=customer["name"],
+            food=customer["food"]
+        )
+        for customer in customers
+    ]
+
+    hall = CinemaHall(hall_number)
+    cinema_bar = CinemaBar
```
LiudmylaKulinchenko, line 22: You can skip creating a variable here, actually, and just use class in 25th line

app/cinema/hall.py
```diff
@@ -0,0 +1,21 @@
...
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(
+        self,
+        movie_name: str,
+        customers: list["Customer"],
+        cleaning_staff: "Cleaner"
+    ) -> None:
+        print(f'\"{movie_name}\" started in hall number {self.number}.')
```
LiudmylaKulinchenko, line 15: You don't need to use `\` in this case, it will work in the same way, because your outer quotes are different. Fix everywhere throughout the project

LiudmylaKulinchenko (changes requested):

app/cinema/bar.py
```diff
@@ -0,0 +1,8 @@
+from app.people.customer import Customer
+
+
+class CinemaBar:
+
+    @staticmethod
+    def sell_product(product: str, customer: "Customer") -> None:
```
Oleksl888, line 7: You don't need to use quotes on class name here - check in all files and change accordingly

app/cinema/hall.py
```diff
@@ -0,0 +1,21 @@
+from app.people.customer import Customer
+from app.people.cinema_staff import Cleaner
```
Oleksl888, line 2: Is a good practice to sort import alphabetically. You can sort either by package name from where it is imported or by imported object

app/main.py
```diff
@@ -1,6 +1,25 @@
...
-    # write you code here
-    pass
+def cinema_visit(
+    customers: list,
+    hall_number: int,
+    cleaner: str,
+    movie: str
+) -> None:
+    customers = [
+        Customer(
```
Oleksl888, line 14: You can have only one for loop here that will optimize your code. Try to create user and sell product in one iteration

app/main.py
```diff
@@ -1,6 +1,25 @@
...
+            name=customer["name"],
+            food=customer["food"]
+        )
+        for customer in customers
+    ]
+    cinema_bar = CinemaBar
```
Oleksl888, line 20: you can call static method directly on class so you don't have to create an object

app/main.py
Oleksl888: This hasn't been implemented

Oleksl888 (changes requested):

MNDonut (approved):

PR #1272, opened 2024-07-25
Solution

mihavryliuk:


app/main.py
```diff
@@ -1,6 +1,34 @@
-# write your imports here
+from app.people.customer import Customer
```
Oleksl888, line 1: Is a good practice to sort imports alphabetically. You can sort by modules from where they are imported or by imported class name

app/main.py
```diff
@@ -1,6 +1,34 @@
...
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(
+    customers: list[dict],
+    hall_number: int,
+    cleaner: str,
+    movie_name: str
+) -> None:
+    customer_objects = [
```
Oleksl888, line 13: You can create customer, add it to a list and sell customer a product in one iteration thus making your code more efficient

app/main.py
```diff
@@ -1,6 +1,34 @@
...
+        Customer(
+            name=customer["name"],
+            food=customer["food"]
+        )
+        for customer in customers
+    ]
+
+    cinema_bar = CinemaBar()
```
Oleksl888, line 21: No need to create instance of this class, you can call method on class directly

Oleksl888 (changes requested):

MNDonut (approved):

PR #1273, opened 2024-07-25
Resolve

ihorutkin:


app/cinema/hall.py
```diff
@@ -0,0 +1,26 @@
...
+    ) -> str:
+        print(f'"{movie_name}" started in hall number {self.number}.')
+        for person in customers:
+            if isinstance(person, dict):
+                customer = Customer(person["name"], person["food"])
+            elif isinstance(person, Customer):
+                customer = person
+            else:
+                raise TypeError("Each customer must be a dict or "
+                                "Customer instance.")
```
LiudmylaKulinchenko, line 23: `This class should have only one method movie_session, that takes movie_name, customers - list of a customers (Customer instances),` - your method should take only list with Customer instances, you don't need to cover other cases

Just put list[Customer] in you annotations

app/main.py
```diff
@@ -1,6 +1,24 @@
...
 
 
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
+def cinema_visit(
+        customers: list,
+        hall_number: int,
+        cleaner: str,
+        movie_name: str
+) -> None:
     # write you code here
```
LiudmylaKulinchenko, line 14: remove this line

app/main.py
```diff
@@ -1,6 +1,24 @@
...
-    pass
+    for person in customers:
+        customer = Customer(person["name"], person["food"])
+        CinemaBar.sell_product(customer=customer, product=customer.food)
+    cleaner = Cleaner(name=cleaner)
```
LiudmylaKulinchenko, line 18: Would be nice to add a blank line between the loop and variables creation just to separate a logic inside a function a little bit. It will make your code more readable

LiudmylaKulinchenko (changes requested):

app/main.py
ihorutkin: I tried to do this, but, unfortunately, i have an error when I run flake8:  
ANN201 Missing return type annotation for public function

app/cinema/hall.py
ihorutkin: Yeah, I did this ckecking because I had a problem:  TypeError: 'Customer' object is not subscriptable.
But, when I wrote this code: 
```
    def movie_session(
            self,
            movie_name: str,
            customers: list[Customer],
            cleaning_staff: "Cleaner"
    ) -> str:
for person in customers:
    person.watch_movie(movie_name)
```
I have this type of problem: 'dict' object has no attribute 'watch_movie'

ihorutkin: I would realy appreciate if you help mi with this type of problem)))

app/main.py
Oleksl888: this hasn't been added

app/main.py
```diff
@@ -1,6 +1,24 @@
...
+    cinema_hall = CinemaHall(number=hall_number)
+    cinema_hall.movie_session(
+        movie_name=movie_name,
+        customers=customers,
```
Oleksl888, line 22: you should pass a list of customer instances here, not it's a dict of key-value pairs

app/main.py
Oleksl888, line 18: no need to create a variable for this you can create instance of cleaner and pass it to movie_session straight away

Oleksl888 (changes requested):

app/cinema/hall.py
ihorutkin: Ok, I fixed it)

app/main.py
ihorutkin: Fixed it too

app/cinema/hall.py
```diff
@@ -0,0 +1,26 @@
+from app.people.customer import Customer
```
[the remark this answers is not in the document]
oliasymak24: You didn't change this

app/cinema/bar.py
```diff
@@ -0,0 +1,7 @@
+from app.people.customer import Customer
+
+
+class CinemaBar:
+    @staticmethod
+    def sell_product(product: str, customer: "Customer") -> None:
```
oliasymak24, line 6: What about previous comment about quotes?

oliasymak24 (changes requested):
Pay attention to the comments you missed last time

oliasymak24 (approved):
Good job!

PR #1275, opened 2024-07-25
Solution for py-cinema-visit

AndriiHamasa:


app/main.py
```diff
@@ -1,6 +1,21 @@
...
+from app.people.cinema_staff import Cleaner
 
 
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(customers: list,
+                 hall_number: int,
+                 cleaner: str,
+                 movie: str) -> None:
```
LiudmylaKulinchenko, line 10: No need to split into lines if it fits in one (I believe, it fits, since you had that in the first place)

app/cinema/hall.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class CinemaHall:
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(self,
+                      movie_name: str,
+                      customers: list,
+                      cleaning_staff: Cleaner) -> None:
```
LiudmylaKulinchenko, line 11: The better practice is to start each parameter from the next line. Fix that, pleased

app/main.py
```diff
@@ -1,6 +1,21 @@
...
+    customers_inst_list = [
+        Customer(customer["name"], customer["food"]) for customer in customers
```
LiudmylaKulinchenko, line 12: You should move loop to the next line, since it's a common practice to devide it like this in comprehensions

app/main.py
```diff
@@ -1,6 +1,21 @@
...
+    ]
+    cineme_hall_inst = CinemaHall(hall_number)
+    cinema_bar_inst = CinemaBar()
```
LiudmylaKulinchenko, line 15: No need to create a class instance to call a static method

app/main.py
```diff
@@ -1,6 +1,21 @@
...
 
 
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(customers: list,
+                 hall_number: int,
+                 cleaner: str,
+                 movie: str) -> None:
+    customers_inst_list = [
```
LiudmylaKulinchenko, line 11: Please, use full words in variable names. It's most common practice in python. You can skip `inst` in your variable name, it will make it shorter and still self-descriptive

LiudmylaKulinchenko (changes requested):

app/main.py
AndriiHamasa: It's over 79 chars (86)

oliasymak24 (approved):
Good job!

PR #1279, opened 2024-07-25
Sollution

Alexandr-Politov:


app/cinema/bar.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class CinemaBar:
+
+    @staticmethod
+    def sell_product(product: str, customer: Customer) -> None:
+        print(f"Cinema bar sold {product} to {customer.name}.")
+
+
+# if __name__ == "__main__":
```
oliasymak24, line 11: Delete these unnecessary comments, in other places too

app/cinema/hall.py
```diff
@@ -0,0 +1,37 @@
...
+        self.number = number
+
+    def movie_session(
+            self,
+            movie_name: str,
+            customers: list[Customer],
+            cleaning_staff: Cleaner
+    ) -> None:
+        print(f'"{movie_name}" started in hall number {self.number}.')
+        for customer in customers:
```
oliasymak24, line 17: for better readability of the code, you should separate the loop from both sides with empty lines, check in other parts of the code as well

app/main.py
```diff
@@ -1,6 +1,24 @@
...
+        hall_number: int,
+        cleaner: str,
+        movie: str
+) -> None:
+    customer_instances = [
+        Customer(customer["name"], customer["food"])
+        for customer in customers
+    ]
+    hall_instance = CinemaHall(hall_number)
+    bar_instance = CinemaBar()
```
oliasymak24, line 18: No need to create an instance to call static method

oliasymak24 (changes requested):

app/main.py
Alexandr-Politov: Agry. But this is a part of a task:
This function should make `Customers` instances, **instance of** `CinemaHall`
and **`CinemaBar`**, instance of `Cleaner`.

MNDonut (approved):

PR #1283, opened 2024-07-26
Solution

pashawarganov:


app/cinema/hall.py
```diff
@@ -0,0 +1,21 @@
...
+        self.number = number
+
+    def movie_session(
+            self,
+            movie_name: str,
+            customers: list[Customer],
+            cleaning_staff: Cleaner
+    ) -> None:
+        print(f'"{movie_name}" started in hall number {self.number}.')
+        for customer in customers:
```
oliasymak24, line 18: for better readability of the code, you should separate the loop from both sides with empty lines, check in other parts of the code as well

app/main.py
```diff
@@ -1,6 +1,23 @@
...
+        hall_number: int,
+        cleaner: str,
+        movie: str
+) -> None:
+    customers = [
+        Customer(customer["name"], customer["food"])
+        for customer in customers
+    ]
+    hall = CinemaHall(hall_number)
+    cb = CinemaBar()
```
oliasymak24, line 18: It is not good practice to create variable name like this, when you will use it in other places of code, it will be not clear what `cb` means

oliasymak24: No need to create a class instance to call a static method

oliasymak24 (changes requested):

app/main.py
pashawarganov: I created instance of CinemaBar, because this is instruction in README. Now I have deleted it, but I think it is necessary to correct this moment in README.

MNDonut (approved):

PR #1284, opened 2024-07-26
implement

ValeriiaKyr:


app/cinema/hall.py
```diff
@@ -0,0 +1,20 @@
+from app.people.cinema_staff import Cleaner
+from app.people.customer import Customer
+
+
+class CinemaHall:
+    def __init__(
+            self,
+            number: int) -> None:
```
MNDonut, line 8: the closing bracket should be always on a separate line, fix everywhere

app/people/customer.py
```diff
@@ -0,0 +1,9 @@
+class Customer:
+    def __init__(self,
+                 name: str,
+                 food: str) -> None:
```
MNDonut, line 4: parameters should be on separate lines and the closing bracket as well

MNDonut (changes requested):

MNDonut (approved):

PR #1285, opened 2024-07-26
Solution

olena4reunited:


app/cinema/bar.py
```diff
@@ -0,0 +1,8 @@
+from __future__ import annotations
+from app.people.customer import Customer
```
MNDonut, line 2: The best practice is to split imports on 3 groups by 1 blank line:
1. default python modules
2. installed modules
3. modules inside your project
Let's fix it everywhere :)

app/main.py
```diff
@@ -1,6 +1,34 @@
...
+
+    for customer in customer_instances:
+        bar_instance.sell_product(
+            customer=customer,
+            product=customer.food
+        )
+
+    cinema_hall_instance.movie_session(
+        movie, customer_instances,
+        cleaner_instance
```
MNDonut, line 33: It's not recommended to place parameters in this way. They should be either in one line or on separate lines like you do in 26-29 rows

app/main.py
```diff
@@ -1,6 +1,34 @@
...
+    customer_instances = [
+        Customer(name=customer["name"], food=customer["food"])
+        for customer in customers
+    ]
+
+    cleaner_instance = Cleaner(cleaner)
+
+    cinema_hall_instance = CinemaHall(hall_number)
+
+    bar_instance = CinemaBar()
```
MNDonut, line 23: You're gonna call sell_product method that is static, thus you don't need instance to call this method

MNDonut (changes requested):

app/main.py
```diff
@@ -1,6 +1,34 @@
...
-    # write you code here
-    pass
+
+def cinema_visit(
+        customers: list,
+        hall_number: int,
+        cleaner: str,
+        movie: str
+) -> None:
+    customer_instances = [
```
Oleksl888, line 15: You can avoid having two iterations by creating customer instances, selling them product and adding to the list in one for-loop. This will make your code more efficient

app/main.py
```diff
@@ -1,6 +1,34 @@
...
+        Customer(name=customer["name"], food=customer["food"])
+        for customer in customers
+    ]
+
+    cleaner_instance = Cleaner(cleaner)
```
Oleksl888, line 20: You can simply pass cleaner instance without storing it in a variable in movie_session method

app/main.py
```diff
@@ -1,6 +1,34 @@
-# write your imports here
+from __future__ import annotations
 
+from app.cinema.bar import CinemaBar
```
Oleksl888, line 3: It is also a nice habit to have imports arranged alphabetically. For instance here you can arrange them by packages from which they have been imported or by names of imported objects

app/cinema/hall.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class CinemaHall:
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(
+            self,
+            movie_name: str,
+            customers: List[Customer],
```
Oleksl888, line 15: As of Python 3.9 you don't need to import `List` type but simply use built-in `list` object for annotations

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,6 +1,39 @@
...
+            ),
+            product=Customer(
+                name=customer["name"],
+                food=customer["food"]
+            ).food
+        )
+
+    cinema_hall_instance.movie_session(
+        movie,
+        [
```
Oleksl888, line 31: Almost there. You still have two iterations. See if you can adjust first for-loop so that you save customer in a variable, call sell_product on it and then add customer to a list. Then when you exit for-loop you can pass that list with customers to movie_session method

Oleksl888 (changes requested):

Oleksl888 (approved):
Good Job!

PR #1286, opened 2024-07-26
Solution

dpiuro:


app/cinema/bar.py
```diff
@@ -0,0 +1,7 @@
+from app.people.customer import Customer
+
+
+class CinemaBar:
+    @staticmethod
+    def sell_product(product: str, customer: "Customer") -> None:
```
Oleksl888, line 6: you don't need to put class name in double quotes. Check in other files as well

app/cinema/hall.py
```diff
@@ -0,0 +1,18 @@
+from app.people.cinema_staff import Cleaner
+
+
+class CinemaHall:
+
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(self, movie_name: str,
+                      customers: list,
```
Oleksl888, line 10: what about data type expected inside of the list?

app/main.py
```diff
@@ -1,6 +1,26 @@
...
+from app.people.cinema_staff import Cleaner
 
 
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(customers: list,
+                 hall_number: int,
+                 cleaner: str,
+                 movie: str) -> list:
```
Oleksl888, line 10: does this function return a list?

app/main.py
```diff
@@ -1,6 +1,26 @@
...
+
+    customers_list = [
```
Oleksl888, line 12: You can avoid having two iterations by creating customer instances, selling them product and adding to the list in one for-loop. This will make your code more efficient

app/main.py
```diff
@@ -1,6 +1,26 @@
...
+        Customer(name=cust["name"], food=cust["food"]) for cust in customers
+    ]
+
+    cinema_bar = CinemaBar()
```
Oleksl888, line 16: you can call static method directly on class without creating an instance

app/main.py
```diff
@@ -1,6 +1,26 @@
...
+    cinema_hall = CinemaHall(number=hall_number)
+
+    for customer in customers_list:
+        cinema_bar.sell_product(product=customer.food, customer=customer)
+
+    cleaning = Cleaner(name=cleaner)
```
Oleksl888, line 22: you don't need to store cleaner in a variable since you not going to reuse it

app/people/customer.py
```diff
@@ -0,0 +1,7 @@
+class Customer:
+    def __init__(self, name: str, food: str) -> None:
+        self.name = name
+        self.food = food
+
+    def watch_movie(self, movie: str) -> None:
+        print(f'{self.name} is watching \"{movie}\".')
```
Oleksl888, line 7: You don't need to use escape symbols here

Oleksl888 (changes requested):

app/cinema/hall.py
```diff
@@ -0,0 +1,19 @@
+from app.people.cinema_staff import Cleaner
+from app.people.customer import Customer
+
+
+class CinemaHall:
+
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(self, movie_name: str,
```
Oleksl888, line 10: It is a good practice to have each parameter starting at new line

app/main.py
Oleksl888: This hasn't been changed

app/main.py
```diff
@@ -1,6 +1,23 @@
...
-    # write you code here
-    pass
+def cinema_visit(customers: list,
+                 hall_number: int,
+                 cleaner: str,
+                 movie: str) -> None:
+
+    cinema_bar = CinemaBar()
+    customers_list = []
+    for cust in customers:
```
Oleksl888, line 14: `cust` is not a meaningful name for a variable you could use `customer_data` instead or something similar that indicates what this variable holds

Oleksl888 (changes requested):

app/cinema/hall.py
```diff
@@ -0,0 +1,20 @@
...
+
+class CinemaHall:
+
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(self,
+                      movie_name: str,
+                      customers: list[Customer],
+                      cleaning_staff: Cleaner) -> str:
```
Oleksl888, line 13: does this method return a string?

app/main.py
Oleksl888: you still have two iterations. can you optimize this code as per comments before?

Oleksl888 (changes requested):

oliasymak24 (approved):

PR #1287, opened 2024-07-26
'Solution'

ABaidin:


app/main.py
```diff
@@ -1,6 +1,24 @@
...
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(
+        customers: list,
+        hall_number: int,
+        cleaner: str,
+        movie: str
+) -> None:
+    cinema_bar = CinemaBar()
```
Oleksl888, line 13: you can call static method directly on class without creating an instance

app/main.py
```diff
@@ -1,6 +1,24 @@
...
+    cinema_hall = CinemaHall(hall_number)
+    cleaner = Cleaner(cleaner)
```
Oleksl888, line 15: you don't need to store cleaner in a variable but create it inside of movie_session method

app/main.py
```diff
@@ -1,6 +1,24 @@
...
+    customers_instances = [
```
Oleksl888, line 16: You can make your code more efficient if you keep only one iteration where you will create customer instances, invoke sell_product on them and add to the final list

app/main.py
```diff
@@ -1,6 +1,24 @@
-# write your imports here
+from app.people.cinema_staff import Cleaner
```
Oleksl888, line 1: Is a good practice to organize imports alphabetically. Here you can arrange them by packages from which they have been imported or by names of imported objects

Oleksl888 (changes requested):

app/main.py
ABaidin: Yes, I agree, but I was following the task. 
It says I need to create an instance of CinemaBar inside function "cinema_visit". 

Do I need to change the code as you say?

app/main.py
ABaidin: Task says cinema_visit function should make instance of Cleaner, 
and movie_session, should take cleaning_staff - cleaner (Cleaner instance). 

Do I need to change the code as you say?

MNDonut (approved):

PR #1288, opened 2024-07-26
Solution

Sergi0bbb:


app/main.py
```diff
@@ -1,6 +1,25 @@
...
-    # write you code here
-    pass
+def cinema_visit(
+        customers: list,
+        hall_number: int,
+        cleaner: str,
+        movie: str
+) -> None:
+    customer_instances = []
+    for i in customers:
```
Oleksl888, line 14: You should give meaningful names to variables

app/main.py
```diff
@@ -1,6 +1,25 @@
...
+        customer = Customer(name=i["name"], food=i["food"])
```
Oleksl888, line 15: Also you have two for-loops in your code. You can make your code more efficient if you keep only one iteration where you will create customer instances, invoke sell_product on them and add to the final list

app/main.py
```diff
@@ -1,6 +1,25 @@
...
+        customer_instances.append(customer)
+
+    bar_instance = CinemaBar()
```
Oleksl888, line 18: you can call static method directly on class without creating an instance

app/main.py
```diff
@@ -1,6 +1,25 @@
...
+    hall_instance = CinemaHall(hall_number)
+    clean_instance = Cleaner(cleaner)
```
Oleksl888, line 20: you don't need to store cleaner in a variable but create it inside of movie_session method

Oleksl888 (changes requested):

Oleksl888 (approved):
Well done!

PR #1290, opened 2024-07-26
Solution

mnezbrytska:


app/cinema/hall.py
```diff
@@ -0,0 +1,18 @@
+from app.people.cinema_staff import Cleaner
+
+
+class CinemaHall:
+
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(self,
```
Oleksl888, line 9: Bad formatting

app/cinema/hall.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class CinemaHall:
+
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(self,
+                      movie_name: str,
+                      customers: list,
```
Oleksl888, line 11: Can you specify also the data types inside of `customers` list?

app/main.py
```diff
@@ -1,6 +1,23 @@
...
-    pass
+def cinema_visit(
+        customers: list,
+        hall_number: int,
+        cleaner: str,
+        movie: str
+) -> None:
+    cleaner_instance = Cleaner(cleaner)
+    hall = CinemaHall(hall_number)
+    customer_instances = [
```
Oleksl888, line 15: You can optimize code here so you have only one iteration that creates customer instances, sells them product and adds to the list. Also you don't need to store Cleaner and CinemaHall in a variable since you are going to reuse these instances

app/main.py
```diff
@@ -1,6 +1,23 @@
...
+        Customer(cust["name"], cust["food"])
+        for cust in customers
```
Oleksl888, line 17: `cust` is not a meaningful name for a variable

app/main.py
```diff
@@ -1,6 +1,23 @@
...
+    ]
+    bar_cinema = CinemaBar()
```
Oleksl888, line 19: You may call `sell_product` method directly on a class without instantiating it

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1,6 +1,26 @@
...
+from app.cinema.hall import CinemaHall
+from app.people.cinema_staff import Cleaner
+from app.people.customer import Customer
 
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+
+def cinema_visit(
+        customers: List[Customer],
```
Oleksl888, line 10: wrong typing here - list of customers is expected in hall.py

app/main.py
```diff
@@ -1,6 +1,26 @@
...
+        hall_number: int,
+        cleaner: str,
+        movie: str
+) -> None:
+    customer_instances = [
+        Customer(customer["name"], customer["food"])
+        for customer in customers
+    ]
+
+    for customer in customer_instances:
```
Oleksl888, line 20: you can create customer instances here instead of having a separate for-loop on line 15. This will make your code more efficient

Oleksl888 (changes requested):

app/main.py
Oleksl888: This hasn't been changed

app/main.py
Oleksl888: Yes you are right, but when you create this list of instances you should also call `sell_product` on each instance in the same for-loop. So my point is to remove the second for-loop.

Oleksl888 (changes requested):

mnezbrytska:
@LiudmylaKulinchenko

Oleksl888 (approved):
Good job! :clap:

PR #1292, opened 2024-07-26
Solution

Bogdan-Rylov:


app/main.py
```diff
@@ -1,6 +1,17 @@
...
-    # write you code here
-    pass
+def cinema_visit(
+        customers: list,
+        hall_number: int,
+        cleaner: str,
+        movie: str
+) -> None:
+    cinema_hall = CinemaHall(hall_number)
+    customers_objects = [Customer(c["name"], c["food"]) for c in customers]
```
Oleksl888, line 14: You can optimize code here so you have only one iteration that creates customer instances, sells them product and adds to the list. Also `c` is not a  meaningful name for a variable

Oleksl888 (changes requested):

MNDonut (approved):

PR #1293, opened 2024-07-28
implement py-cinema-visit

Maksym-Turenko:


app/main.py
```diff
@@ -1,6 +1,28 @@
...
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(
+        customers: list[dict],
+        hall_number: int,
+        cleaner: str,
+        movie: str
+) -> None:
+    customer_instances = [
```
Oleksl888, line 13: You can rewrite this so you have only one iteration that creates customer instances, sells them product and adds to the list. This will make your code more efficient

app/main.py
```diff
@@ -1,6 +1,28 @@
...
+        Customer(
+            name=customer["name"],
+            food=customer["food"]
+        ) for customer in customers]
+
+    cinema_bar = CinemaBar()
```
Oleksl888, line 19: No need to create an instance of CinemaBar - you can call staticmethod directly on a class. ( I know it said to do so in requirements but we are looking at more optimal solution here)

app/main.py
```diff
@@ -1,6 +1,28 @@
...
+    for customer in customer_instances:
+        cinema_bar.sell_product(product=customer.food, customer=customer)
+    cinema_hall = CinemaHall(number=hall_number)
```
Oleksl888, line 22: you can omit storing hall and cleaner instances in a variable since we are not going to reuse them

Oleksl888 (changes requested):

Oleksl888 (approved):
Well done!

PR #1307, opened 2024-08-07
Solution

CheshireKate:


app/cinema/bar.py
```diff
@@ -0,0 +1,8 @@
+from app.people import customer as client
```
Dimosphen1, line 1: There is no need to redefine `Customer` class name with `client` here, use `Customer` class directly here and in the other places as well. Do not place class inside `""`. Use `from __future__ import annotations` import in case you experience issues with type annotation.

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1,6 +1,20 @@
...
+from app.people.cinema_staff import Cleaner
+from app.cinema.bar import CinemaBar
+from app.cinema.hall import CinemaHall
 
 
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(
+        customers: list, hall_number: int, cleaner: str, movie: str
```
```suggestion
        customers: list[dict], hall_number: int, cleaner: str, movie: str
```

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #1312, opened 2024-08-09
Solution

bdnnk:


app/cinema/bar.py
```diff
@@ -0,0 +1,7 @@
+from app.people.customer import Customer
+
+
+class CinemaBar:
+    @staticmethod
+    def sell_product(customer: Customer, product: Customer) -> None:
```
pavlopro, line 6: Correct annotation

app/main.py
```diff
@@ -1,6 +1,23 @@
...
-    # write you code here
-    pass
+def cinema_visit(customers: list, hall_number: int,
+                 cleaner: str, movie: str) -> None:
+
+    customer_objects = [Customer(name=customer["name"],
+                                 food=customer["food"])
+                        for customer in customers]
+
+    cinema_bar = CinemaBar()
```
pavlopro, line 14: No need to have this variable `cinema_bar`. Use class instead as `sell_product` method is a staticmethod

pavlopro (changes requested):

app/cinema/bar.py
pavlopro: This was not resolved

pavlopro (changes requested):

bdnnk:
Please explain it to me. Are you demanding that I change the annotation type because it doesn't work in version 3.10 and below? Could I, for example, just import annotations from the __future__ module and leave it as it is?

app/cinema/bar.py
```diff
@@ -0,0 +1,4 @@
+class CinemaBar:
+    @staticmethod
+    def sell_product(customer: str, product: str) -> None:
```
pavlopro, line 3: `def sell_product(customer: Customer, product: str) -> None:`

pavlopro (changes requested):

pavlopro (approved):
Good Job!

PR #1320, opened 2024-08-14
My solution

lavryniuk-dev:


app/cinema/hall.py
```diff
@@ -0,0 +1,20 @@
...
+
+class CinemaHall:
+
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(self,
+                      movie_name: str,
+                      customers: list[Customer],
+                      cleaning_staff: Cleaner) -> None:
```
LiudmylaKulinchenko, line 13: It's ok, but more readable and common to separate each parameter in one line
Please, fix throughout your code

app/main.py
```diff
@@ -1,6 +1,20 @@
...
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(customers: list,
+                 hall_number: int,
+                 cleaner: str,
+                 movie: str) -> None:
+    customer_objects = [Customer(customer["name"],
+                                 customer["food"])
+                        for customer in customers]
```
LiudmylaKulinchenko, line 13: Same for list comprehension: let's move to the next line the code, that you have after [
Same with the code that you have before ]

LiudmylaKulinchenko (changes requested):
Your solution is good, but let's fix formatting!

LiudmylaKulinchenko (approved):
Good job

PR #1322, opened 2024-08-16
Solution

ip-ilya:


app/main.py
```diff
@@ -1,6 +1,34 @@
-# write your imports here
+from app.people.cinema_staff import Cleaner
+from app.people.customer import Customer
```
LLkaia, line 3: No reason for blank line between custom imports

app/main.py
```diff
@@ -1,6 +1,34 @@
...
+        hall_number: int,
+        cleaner: str,
+        movie: str
+) -> None:
+
+    cleaner_instance = Cleaner(name=cleaner)
+    customer_instances = [
+        Customer(name=customer["name"], food=customer["food"])
+        for customer in customers
+    ]
```
LLkaia, line 19: You have iteration through customers on line 16 and 24. If you create cinema bar, and then iterate through customers, create Customer instance and then sell food for them under same loop, it will help you to avoid looping twice.

LLkaia (changes requested):
Your test failed + additional comments

LLkaia (approved):
Great job!
Just comment for consideration, it's not necessary to use backslash in strings
If you string defined with `"  "`, you're free to use `'  '` in it.  If you string defined with `'   '`, you're free to use `"   "` in it without backslash.
So, just  `print(f'"{movie_name}" ended.')`

PR #1336, opened 2024-08-26
Solution

tetianasobko:


app/cinema/hall.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class CinemaHall:
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(
+            self, movie_name: str, customers: list[Customer], cleaner: Cleaner
+    ) -> None:
+        print(f'\"{movie_name}\" started in hall number {self.number}.')
```
LLkaia, line 12: No reason use backslashes here cause " used in '

app/cinema/hall.py
```diff
@@ -0,0 +1,19 @@
...
+
+        for customer in customers:
+            customer.watch_movie(movie_name)
+
+        print(f'\"{movie_name}\" ended.')
```
LLkaia, line 17: Same here

LLkaia (changes requested):

PR #1353, opened 2024-09-11
Solution

mshmygel:


app/main.py
```diff
@@ -1,6 +1,41 @@
...
+hall_number = 5
+cleaner_name = "Anna"
+movie = "Madagascar"
+
+cinema_visit(
+    customer_list=customers,
+    hall_num=hall_number,
+    cleaner_person=cleaner_name,
+    movie_title=movie
+)
```
AnyoneClown, line 41: Remove the unnecessary code you used for example

app/main.py
```diff
@@ -1,6 +1,41 @@
...
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(
+        customer_list: list,
+        hall_num: int,
+        cleaner_person: str,
+        movie_title: str
+) -> None:
+    cb = CinemaBar()
```
AnyoneClown, line 13: bad variable name

AnyoneClown (changes requested):

app/main.py
mshmygel: deleted it

app/main.py
mshmygel: changed it

AnyoneClown (approved):
Good Job!

PR #1358, opened 2024-09-13
'Solution'

KKvladys:


app/cinema/hall.py
```diff
@@ -0,0 +1,20 @@
...
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(
+            self,
+            movie_name: str,
+            customers: list[Customer],
+            cleaning_staff: Cleaner
+    ) -> None:
+        print(f'\"{movie_name}\" started in hall number {self.number}.')
```
AnyoneClown, line 16: Use `f' "Some string" '` if you need to use `”` inside a string instead of `f' \” Some string \” '`




</body>
</html>

AnyoneClown (changes requested):

AnyoneClown (approved):
LGTM!

PR #1371, opened 2024-09-20
Completed task

Mxbely:
Created "people" and "cinema" package, then created next moduls: bar.py, cinema_staff.py, customer.py, hall.py

### DIFF
--- a/app/cinema/hall.py
+++ b/app/cinema/hall.py
@@ -1,0 +1,17 @@
+from app.people.customer import Customer
+from app.people.cinema_staff import Cleaner
+# from time import time
+
+
+class CinemaHall:
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(self, movie_name: str,
+                      customers: list[Customer],
+                      cleaning_staff: Cleaner) -> None:
+        print(f"\"{movie_name}\" started in hall number {self.number}.")
+        for customer in customers:
+            customer.watch_movie(movie=movie_name)
+        print(f"\"{movie_name}\" ended.")
+        cleaning_staff.clean_hall(hall_number=self.number)
--- a/app/main.py
+++ b/app/main.py
@@ -1,6 +1,23 @@
-# write your imports here
+from app.cinema.bar import CinemaBar
+from app.cinema.hall import CinemaHall
+from app.people.customer import Customer
+from app.people.cinema_staff import Cleaner
 
 
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(
+        customers: list,
+        hall_number: int,
+        cleaner: str,
+        movie: str) -> None:
+    customer_list = [
+        Customer(
+            name=customer["name"],
+            food=customer["food"]
+        )
+        for customer in customers
+    ]
+    for customer in customer_list:
+        CinemaBar.sell_product(customer.food, customer)
+    hall = CinemaHall(hall_number)
+    cleaner_name = Cleaner(cleaner)
+    hall.movie_session(movie, customer_list, cleaner_name)
--- a/app/cinema/bar.py
+++ b/app/cinema/bar.py
@@ -1,0 +1,7 @@
+from app.people.customer import Customer
+
+
+class CinemaBar:
+    @staticmethod
+    def sell_product(product: str, customer: Customer) -> None:
+        print(f"Cinema bar sold {product} to {customer.name}.")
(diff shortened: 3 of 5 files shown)


dmytrik (approved):
good job!

app/cinema/hall.py
```diff
@@ -0,0 +1,17 @@
+from app.people.customer import Customer
+from app.people.cinema_staff import Cleaner
+# from time import time
```
AnyoneClown, line 3: Delete unused import

app/cinema/hall.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class CinemaHall:
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(self, movie_name: str,
+                      customers: list[Customer],
+                      cleaning_staff: Cleaner) -> None:
+        print(f'\"{movie_name}\" started in hall number {self.number}.')
```
AnyoneClown, line 13: Use `f’ “Some string” ‘` if you need to use `”` inside a string instead of `f” \” Some string \” “`

app/main.py
```diff
@@ -1,6 +1,23 @@
...
+            name=customer["name"],
+            food=customer["food"]
+        )
+        for customer in customers
+    ]
+    for customer in customer_list:
+        CinemaBar.sell_product(customer.food, customer)
+    hall = CinemaHall(hall_number)
+    cleaner_name = Cleaner(cleaner)
+    hall.movie_session(movie, customer_list, cleaner_name)
```
AnyoneClown, line 23: You can immediately drop a method from this class, without first initialising it in variable `hall`

```
CinemaHall(hall_number).movie_session()
```

AnyoneClown (changes requested):

app/main.py
```diff
@@ -1,6 +1,23 @@
...
+        movie: str) -> None:
+    customer_list = [
+        Customer(
+            name=customer["name"],
+            food=customer["food"]
+        )
+        for customer in customers
+    ]
+    for customer in customer_list:
+        CinemaBar.sell_product(customer.food, customer)
```
AnyoneClown, line 20: You can optimise the code using only 1 loop

Mxbely: i can write smth like `[CinemaBar.sell_product(customer.food, customer) for customer in [Customer(name=customer["name"],food=customer["food"]) for customer in customers]]
` 
but when i call `.movie_session` i use parameter `customer_list`, 
if i didn't save `customer_list`, i will be need calculate this again

AnyoneClown: We can simply do it inside of 1 `for loop`, as you already figured out

app/main.py
```diff
@@ -9,15 +9,10 @@ def cinema_visit(
...
-            food=customer["food"]
-        )
-        for customer in customers
-    ]
-    for customer in customer_list:
-        CinemaBar.sell_product(customer.food, customer)
-    hall = CinemaHall(hall_number)
+    customer_list = []
+    for customer in customers:
+        custom = Customer(name=customer["name"], food=customer["food"])
```
AnyoneClown, line 14: Use more descriptive variable name, for example `customer_instance` or `visitor`

app/cinema/hall.py
```diff
@@ -10,8 +9,8 @@ def __init__(self, number: int) -> None:
     def movie_session(self, movie_name: str,
                       customers: list[Customer],
                       cleaning_staff: Cleaner) -> None:
```
AnyoneClown, line 11: Bad identation, let's start all arguments from new line. For example:

```
def movie_session(
    self,
    movie_name: str
) -> None:
```

app/main.py
```diff
@@ -9,15 +9,10 @@ def cinema_visit(
...
+        customer_list.append(custom)
+        CinemaBar.sell_product(custom.food, custom)
     cleaner_name = Cleaner(cleaner)
```
AnyoneClown, line 17: You fixed CinemaHall, but you should check all cases of that problem. For example, you can also you Cleaner instance without declaration it into a variable

AnyoneClown (changes requested):

app/main.py
```diff
@@ -1,6 +1,21 @@
...
+from app.cinema.hall import CinemaHall
+from app.people.customer import Customer
+from app.people.cinema_staff import Cleaner
 
 
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(
+        customers: list,
```
Nattalli, line 8: list of integers, strings? specify

Nattalli (approved):
well done but one small comment left

PR #1377, opened 2024-09-27
Solution

VladyslavBon:


app/main.py
```diff
@@ -1,6 +1,16 @@
...
+from app.people.customer import Customer
+from app.people.cinema_staff import Cleaner
 
 
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(
+        customers: list, hall_number: int, cleaner: str, movie: str
+) -> None:
```
Mxbely, line 9: it is bad practice, rewrite the function parameters, for example:
```
def cinema_visit(
        customers: list,
        hall_number: int,
) -> None:
```

app/main.py
```diff
@@ -1,6 +1,16 @@
...
+    cinema_bar = CinemaBar()
+    cinema_hall = CinemaHall(hall_number)
+    cinema_customers = [Customer(**customer) for customer in customers]
+    cinema_cleaner = Cleaner(cleaner)
+    for cinema_customer in cinema_customers:
+        cinema_bar.sell_product(cinema_customer.food, cinema_customer)
+    cinema_hall.movie_session(movie, cinema_customers, cinema_cleaner)
```
Mxbely, line 16: you can call class method without using class instance, for example:
`CinemaBar.sell_product( "some params" )`
and i recomend rewrite this code without instances classes

Mxbely (changes requested):

AnyoneClown (approved):
Good job!

PR #1380, opened 2024-09-27
Solution

InsarovKuzma:


app/cinema/hall.py
```diff
@@ -0,0 +1,15 @@
+from app.people.customer import Customer
+from app.people.cinema_staff import Cleaner
```
YukihiroSM, line 2: Why do you need these imports if you are not using them?

app/main.py
```diff
@@ -1,6 +1,19 @@
...
+from app.cinema.bar import CinemaBar
+from app.cinema.hall import CinemaHall
+from app.people.cinema_staff import Cleaner
+from app.people.customer import Customer
 
 
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(customers: list, hall_number: int,
```
YukihiroSM, line 8: It's better to specify list of what do you expect here. For example: list[int], or list[dict]

YukihiroSM (changes requested):

app/cinema/hall.py
InsarovKuzma: Because i use them in fucntion annotaition. Have flake8 F821 undefined name

InsarovKuzma: > Why do you need these imports if you are not using them?

Because i use them in fucntion annotaition. Have flake8 F821 undefined name

YukihiroSM: Then it's better to remove double quotes

app/cinema/bar.py
```diff
@@ -0,0 +1,7 @@
+from app.people.customer import Customer
+
+
+class CinemaBar:
+    @staticmethod
+    def sell_product(product: str, customer: "Customer") -> None:
```
YukihiroSM, line 6: Same as in hall.py here. Fix it, please.

YukihiroSM (changes requested):

YukihiroSM (approved):
Good job!

PR #1382, opened 2024-09-29
implement Cinema

94nj111:


skyfoxwork (approved):
I think it is good!)
But in "def cinema_visit" I would use list comprehension:

    customers_list = [
        Customer(customer.get("name"), customer.get("food"))
        for customer in customers
    ]

instead of

    customers_list = []
    for customer in customers:
        customers_list.append(Customer(customer["name"], customer["food"]))

and used 

customer.get("name"), customer.get("food")

instead of

customer["name"], customer["food"]

and delated "from __future__ import annotations"
I think еhis is not necessary.

app/main.py
```diff
@@ -1,6 +1,20 @@
...
+                 hall_number: int,
+                 cleaner: str,
+                 movie: str) -> None:
+    customers_list = [
+        Customer(customer["name"], customer["food"])
+        for customer in customers
+    ]
+    cleaner_obj = Cleaner(cleaner)
+    hall_obj = CinemaHall(hall_number)
+    bar_obj = CinemaBar()
```
AnyoneClown, line 17: We can use `CinemaBar` methods even without variable. Example:

```
CinemaBar.sell_product()
```

AnyoneClown (changes requested):

PR #1384, opened 2024-09-29
Solution

vladyslav-tmf:


app/cinema/bar.py
```diff
@@ -0,0 +1,7 @@
+from app.people.customer import Customer
+
+
+class CinemaBar:
+    @staticmethod
+    def sell_product(product: str, customer: "Customer") -> None:
```
94nj111, line 6: This is the first time I have seen such an annotation, how does it work? Maybe it would be better to use __future annotation__?

94nj111 (approved):
lgtm!

app/cinema/bar.py
vladyslav-tmf: > This is the first time I have seen such an annotation, how does it work? Maybe it would be better to use **future annotation**?

I used this to avoid a syntax error because the Customer class is not yet defined when annotation is initiated.
**Future annotations** are useful, but I chose standard-type hints for clarity.

AnyoneClown: Why will there be a syntax error? We already import this class, so it is implemented. Let me remind you that the `type hint` has almost no effect on the interpreter, we simply specify the type we expect to receive.

Fix it in all cases

app/main.py
```diff
@@ -1,6 +1,23 @@
...
 
 
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(customers: list,
+                 hall_number: int,
+                 cleaner: str,
+                 movie: str) -> None:
+    cinema_bar = CinemaBar()
```
AnyoneClown, line 11: If we directly use `CinemaBar()`, but do not pass any arguments to `__init__`, we can immediately use it in the code. For example:

```
CinemaBar().sell_product(customer.food, customer)
```

AnyoneClown (changes requested):

PR #1386, opened 2024-09-30
implement py-cinema-visit

yuliiapyrih:


app/main.py
```diff
@@ -1,6 +1,24 @@
...
-    # write you code here
-    pass
+def cinema_visit(
+        customers: list,
+        hall_number: int,
+        cleaner: str,
+        movie: str
+) -> None:
+    customers_list = []
+    cinema_bar = CinemaBar()
```
LiudmylaKulinchenko, line 14: No need to create a class instance to call a static method

LiudmylaKulinchenko (changes requested):

app/main.py
yuliiapyrih: Done

Nattalli (approved):

PR #1397, opened 2024-10-04
solution

SerhiiK05:


app/main.py
```diff
@@ -1,6 +1,22 @@
...
-    # write you code here
-    pass
+def cinema_visit(
+        customers: list,
+        hall_number: int,
+        cleaner: str,
+        movie: str
+) -> None:
+    cinema_hall = CinemaHall(hall_number)
+    cinema_bar = CinemaBar()
```
sumseer, line 14: CinemaBar instance is not necessary, since it's a static method

sumseer (changes requested):
Not sure though)

sumseer (approved):
Nice!
(I hope it's right xd)

PR #1412, opened 2024-10-05
Solution

sofiiasavkova:


app/cinema/bar.py
```diff
@@ -0,0 +1,7 @@
+from app.people.customer import Customer
```
Rabb1t-Artem, line 1: you don't need to import the whole class in this file

app/cinema/hall.py
```diff
@@ -0,0 +1,22 @@
+from app.people.customer import Customer
+from app.people.cinema_staff import Cleaner
```
Rabb1t-Artem, line 2: you don't need to import the whole class in this file

Rabb1t-Artem (changes requested):

Arsen-hrynevych (approved):
Ok

PR #1413, opened 2024-10-06
solution

sashasyrota:


app/cinema/hall.py
```diff
@@ -0,0 +1,19 @@
...
+
+    def movie_session(
+            self,
+            movie_name: str,
+            customers: list,
+            cleaning_staff: Cleaner
+    ) -> None:
+        print(f'"{movie_name}" started in hall number {self.number}.')
+        for element in customers:
+            element.watch_movie(movie_name)
```
Dimosphen1, line 17: It is better to use `customer` instead of `element` as a variable name to clearly indicate its content here and in the `app/main.py` file as well

app/cinema/hall.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class CinemaHall:
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(
+            self,
+            movie_name: str,
+            customers: list,
```
Dimosphen1, line 12: Indicate a list of what instances you expect to obtain for `customers` in the type annotation

app/cinema/bar.py
```diff
@@ -0,0 +1,8 @@
+from __future__ import annotations
```
Dimosphen1, line 1: Add one blank line between different import groups to follow PEP-8 here and in the other places as well

Dimosphen1 (changes requested):

pavlopro (approved):
GJ!

PR #1419, opened 2024-10-08
Solution

yulia-pl:


app/main.py
```diff
@@ -1,6 +1,32 @@
...
-    pass
+def cinema_visit(
+        customers: List[Dict[str, str]], hall_number: int,
+        cleaner: str, movie: str
+) -> None:
+    customer_instances = [
+        Customer(name=customer["name"], food=customer["food"])
+        for customer in customers
+    ]
+    cb = CinemaBar()
```
Le0n-K, line 16: good practice: it is better to write the variable in full, for example **cinema_bar** instead of **cb**)
otherwise, good job)

Le0n-K (approved):

PR #1435, opened 2024-10-15
solution

popeye88:


app/main.py
```diff
@@ -1,6 +1,31 @@
...
+            food=customer["food"]
+        ) for customer in customers
+    ]
+
+    hall_instance = CinemaHall(hall_number)
+    bar_instance = CinemaBar()
+    cleaner_instance = Cleaner(cleaner)
+
+    for customer in customer_instances:
+        bar_instance.sell_product(customer=customer, product=customer.food)
```
pavlopro, line 25: No need to have `bar_instance` variable. Use class `CinemaBar` instead as `sell_product` method is a `staticmethod`

pavlopro (changes requested):

pavlopro (approved):
GJ!

PR #1439, opened 2024-10-16
Solution

serg121192:


app/cinema/bar.py
```diff
@@ -0,0 +1,8 @@
+from __future__ import annotations
```
pavlopro, line 1: I suppose you don't need this import

LLkaia (approved):
Good Job!

PR #1441, opened 2024-10-18
cinema

Anatolich911:


app/people/cinema_staff.py
```diff
@@ -0,0 +1,6 @@
+class Cleaner:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def clean_hall(self, hall_number: int) -> None:
+        print(f'Cleaner {self.name} is cleaning hall number {hall_number}.')
```
Alkaponees, line 6: Screenshot from 2024-10-18 17-55-25
Recommend to watch error in checks. Flask give you explanation how you can fix it

Alkaponees (changes requested):
Check error in GitHub check section

app/people/cinema_staff.py
Anatolich911: i should've checked GH actions first. Thanks for pointing it

Alkaponees (approved):
LGMT!

PR #1495, opened 2024-11-21
Solution

MarianaAB19:


app/cinema/hall.py
```diff
@@ -0,0 +1,23 @@
...
+from app.people.customer import Customer
+
+
+class CinemaHall:
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(
+            self, movie_name: str,
+            customers: list,
```
vasylhnatiuk, line 11:
```suggestion
            customers: list[dict],
```

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+from app.cinema.hall import CinemaHall
+from app.people.customer import Customer
+from app.people.cinema_staff import Cleaner
 
 
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(
+        customers: list,
```
```suggestion
        customers: list[Customer],
```

app/cinema/hall.py
```diff
@@ -0,0 +1,23 @@
...
+            cleaning_staff: Cleaner
+    ) -> str:
+        print(f'\"{movie_name}\" started in hall number {self.number}.')
+        if isinstance(customers[0], dict):
```
vasylhnatiuk, line 15: why do you need it?

vasylhnatiuk (changes requested):

app/cinema/hall.py
MarianaAB19: Hi!
To check if customers is a list of dictionaries, then if customers is a list of Customer objects.
From tests I understood that both cases are possible, that's why I did this check.
How to do it better?)
Thanks!

PR #1517, opened 2024-12-02
first try

hlibswork:


app/main.py
```diff
@@ -1,6 +1,17 @@
...
+from cinema.bar import CinemaBar
+from cinema.hall import CinemaHall
 
 
 def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+    peoples = []
+    for customer in customers:
+        peoples.append(Customer(customer["name"], customer["food"]))
```
vsmutok, line 11: you can use list comprehension here

app/main.py
```diff
@@ -1,6 +1,17 @@
-# write your imports here
+from __future__ import annotations
```
vsmutok, line 1: I think this import redundant here

vsmutok (changes requested):

app/main.py
hlibswork: Fixed it. Can you tell me what the Assertion error is in the tests? Because I can't figure it out.

vasylhnatiuk (approved):

PR #1520, opened 2024-12-03
record of events in the cinema

valerii-kashpur:
- added bar class
- added hall class
- added customer class
- added cleaner class
- added cinema_visit func wich invoke event loger


dottfmar (approved):
Well done!

PR #1543, opened 2024-12-05
Solution

lurd22:


app/main.py
```diff
@@ -1,6 +1,28 @@
...
+
+    customer_instances = [
+        Customer(name=customer["name"], food=customer["food"])
+        for customer in customers
+    ]
+
+    cinema_bar = CinemaBar()
+
+    for customer in customer_instances:
+        cinema_bar.sell_product(product=customer.food, customer=customer)
```
pavlopro, line 21: sell product is a static method, so no need to have instance of CinemaBar class

pavlopro (changes requested):

pavlopro (approved):
GJ

PR #1555, opened 2024-12-08
Dev

Arty-tech-dotcom:


TretiakHelen (changes requested):
I don't understand why the bar.py file doesn't pass the flake8 check, but in the other files (hall.py, main.py, cinema_staff.py, customer.py) at least there should be "W292 no newline at end of file". Use the "flake8 app" command and fix all the warnings. You can read the error description here "https://www.flake8rules.com"

PR #1636, opened 2025-01-25
Solution

ip-ilya:


app/people/cinema_staff.py
```diff
@@ -0,0 +1,12 @@
...
+    def __init__(
+            self,
+            name: str
+    ) -> None:
+        self.name = name
+
+    def clean_hall(
+            self,
+            hall_number: int
+    ) -> None:
```
vsmutok, line 11:
```suggestion
class Cleaner:
    def __init__(self, name: str) -> None:
        self.name = name

    def clean_hall(self, hall_number: int) -> None:
```
If you can write it in one line, it's better to do so.

PR #1702, opened 2025-02-21
Solution

LLevkovych:


app/cinema/hall.py
```diff
@@ -0,0 +1,14 @@
+class CinemaHall:
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(self, movie_name: str,
+                      customers : list,
+                      cleaning_staff : int
+                      ) -> None:
+
+        print(f""""{movie_name}" started in hall number {self.number}.""")
```
Oneyura, line 10: You can use ' ' if you need use " " in string.
Also, you can use \" \" to use this symbol in string

Oneyura (approved):
Well done, but next time use ' ' or \" \".

PR #1709, opened 2025-02-21
Solution

Lerity12:


app/cinema/bar.py
```diff
@@ -0,0 +1,7 @@
+from app.people.customer import Customer
+
+
+class CinemaBar:
+    @staticmethod
+    def sell_product(product: str, customer: Customer) -> None:
+        print(f'Cinema bar sold {product} to {customer.name}.')
```
Yurkiv313, line 7: print(f"Cinema bar sold {product} to {customer.name}.")

replace ' > "

app/cinema/hall.py
```diff
@@ -0,0 +1,22 @@
...
+
+    def movie_session(
+            self,
+            movie_name: str,
+            customers: list[Customer],
+            cleaner: Cleaner
+    ) -> None:
+        print(f'"{movie_name}" started in '
+              f'hall number {self.number}.')
+
```
Yurkiv313, line 17: ) -> None:
        print(f'"{movie_name}" started in hall number {self.number}.')
        
  make one line

Yurkiv313 (changes requested):
just copy from the comments

app/cinema/bar.py
Lerity12: got it, thanks for the advice

PR #1722, opened 2025-02-24
task solution

yarms1:


Maaax256 (changes requested):
check pytest and flake8

Maaax256 (approved):
good

PR #1741, opened 2025-03-01
task

Matsebon:


app/main.py
```diff
@@ -1,6 +1,28 @@
...
-    pass
+def cinema_visit(
+        customers: list,
+        number: int,
+        cleaner: str,
+        movie: str
+) -> None:
+    customer_objects = [
+        Customer(name=c["name"], food=c["food"])
+        for c in customers
```
vsmutok, line 15: `c` bad variable name

vsmutok (changes requested):

app/people/customer.py
```diff
@@ -0,0 +1,7 @@
+class Customer:
+    def __init__(self, name: str, food: str) -> None:
+        self.name = name
+        self.food = food
+
+    def watch_movie(self, movie: str) -> None:
+        print(f'{self.name} is watching \"{movie}\".')
```
SerhiyTo, line 7: It's better to change this code to `print(f"{self.name} is watching '{movie}'")` because you use double quotes as the main ones in the rest of the code

app/main.py
```diff
@@ -1,6 +1,28 @@
...
+) -> None:
+    customer_objects = [
+        Customer(name=customer_var["name"], food=customer_var["food"])
+        for customer_var in customers
+        if "name" in customer_var and "food" in customer_var
+    ]
+    if not customer_objects:
+        print("No valid customers found!")
+        return
+    for cust in customer_objects:
```
SerhiyTo, line 21: It is better to give an even clearer name to this variable

SerhiyTo (changes requested):

app/people/customer.py
Matsebon: photo_2025-03-12_17-19-37
Thank you! But I fail tests when I change it like that, so this way is more correct I guess. I just changed var "cust" to "customer". Tests passed, flake8 check passed

vsmutok (approved):
Great!

PR #1758, opened 2025-03-11
Solution

lexm7:


lexm7 (commented):
Solution was updated quotes in customer.py

PR #1845, opened 2025-04-29
Solution

Nib20:


app/cinema/hall.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class CinemaHall:
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(
+            self,
+            movie_name: str,
+            customers: List[Customer],
```
MNDonut, line 13: you can use built-it list type for annotation

MNDonut (approved):

PR #1933, opened 2025-06-23
Solution for py-cinema-visit

Hetmarum:


app/people/cinema_staff.py
```diff
@@ -0,0 +1,6 @@
+class Cleaner:
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def clean_hall(self, hall_number: int) -> str:
```
Arch0998, line 5: You've specified the return type as str, but the method doesn't actually return anything - it only uses print(...). This contradicts the -> str annotation.

PR #1935, opened 2025-06-24
'Solution'

DanilHamster:


Orixara (approved):
Great job, you have good one with list comprehension, I didn't do like that, but I like it, I will approve you task, but I suggest you check out my suggestions for changing the style when entering arguments

PR #1949, opened 2025-06-25
Solution

VadymKyrylko:


Gleg0 (approved):
Well done, but you need to change "" to '' and in the '' use "" to wrap movie name (don`t use \"). Than flake8 will don`t throw error.

PR #1956, opened 2025-06-26
Update main.py

B1g-Bear:


app/main.py
```diff
@@ -1,6 +1,18 @@
...
-def cinema_visit(customers: list, hall_number: int, cleaner: str, movie: str):
-    # write you code here
-    pass
+def cinema_visit(
+        customers: list,
+        hall_number: int,
+        cleaner: str,
+        movie: str
+) -> None:
+    customer_objects = [Customer(name=c['name'], food=c['food']) for c in customers]
```
Salmon-eat, line 13: flake8 speaks of a long line

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+    hall = CinemaHall(hall_number=hall_number)
+    cleaning_staff = Cleaner(name=cleaner)
+    for cust in customer_objects:
+        CinemaBar.sell_product(product=cust.food, customer=cust)
+    hall.movie_session(movie_name=movie, customers=customer_objects, cleaning_staff=cleaning_staff)
```
Salmon-eat, line 18: and here is a long one. I think so:)

app/main.py
Salmon-eat: I forgot, you need to fix it with ' --> "

PR #1960, opened 2025-06-26
Solution

bohdankolisnyk04:


reinvvay (approved):
Seems fine to me, but I recommend changing the outer double quotes to single quotes, as suggested by flake8.

PR #2021, opened 2025-07-23
solution

xnquaint:


xnquaint:
customer.py - inside this module create Customer class, its __init__ method takes and stores name, food - food that customer wants to buy in cinema bar. This class should have only one method watch_movie, this method takes movie and prints what movie customer is watching.

PR #2026, opened 2025-07-26
Solution

Konstantin-Svt:


Konstantin-Svt:
@mateacademy-ai-mentor In Documentation README.md it is specified that print output should be in English, not Russian

PR #2079, opened 2025-09-03
cinema: added impl of cinema

kenu21:


kenu21:
@mateacademy-ai-mentor 

> Incorrect parameter order in cinema_visit
> Location: app/main.py, function definition (likely near the top of the file). Look for the line starting with def cinema_visit(.
> What’s wrong: The current signature is (customers, hall_number, cleaner, movie) but the requirement is (movie, customers, hall_number, cleaner).
> Why it matters: This violates the explicit function signature requirement and will likely cause tests or integrations expecting the specified order to fail.
> How to fix: Reorder the parameters in the function definition to def cinema_visit(movie, customers, hall_number, cleaner): and ensure any internal usage (argument unpacking or calls within the file/tests) aligns with this order.

You're wrong. Your suggestion will break existing tests! They are write for current order!!! See tests in PR!!!

kenu21:
@mateacademy-ai-mentor 
You're wrong. Your suggestion will break existing tests! They are write for current order!!! See tests in PR!!!

PR #2100, opened 2025-09-19
Add a solution

Sontsevolosa:


app/cinema/hall.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class CinemaHall:
+    def __init__(self, number: int) -> None:
+        self.number = number
+
+    def movie_session(self, movie_name: str,
+                      customers: list, cleaning_staff: Cleaner) -> None:
+        print('"' + movie_name + '" started in hall number '
+              + str(self.number) + ".")
```
vsmutok, line 11: Use f-strings here instead of string concatenation

app/cinema/hall.py
```diff
@@ -0,0 +1,17 @@
...
+
+        for customer in customers:
+            customer.watch_movie(movie_name)
+
+        print('"' + movie_name + '" ended.')
```
vsmutok, line 16: Use f-strings here instead of string concatenation

app/people/customer.py
```diff
@@ -0,0 +1,7 @@
+class Customer:
+    def __init__(self, name: str, food: str) -> None:
+        self.name = name
+        self.food = food
+
+    def watch_movie(self, movie: str) -> None:
+        print(self.name, 'is watching "' + movie + '".')
```
vsmutok, line 7: Use f-strings here instead of string concatenation

vsmutok (changes requested):

PR #2445, opened 2026-02-08
Solution: implemented cinema visit system

SerhiiHryniuk:


app/main.py
Reverad, line 1: This could be done in one loop, but no big deal since it's already working.

Reverad (approved):

SerhiiHryniuk:
@mateacademy-ai-mentor Hi, I noticed a conflict between the task requirements and the provided tests.
The task description specifies the parameter order as cinema_visit(movie, customers, hall_number, cleaner), but the tests call the function as cinema_visit(customers, hall_number, cleaner, movie) — which means if I revert to the original order, all 3 tests will fail with a TypeError.
I adjusted the parameter order to match the tests since they are the source of truth for the expected behavior. Could you clarify which should take priority — the task description or the tests? If the task description is correct, the test calls would need to be updated accordingly.

PR #2577, opened 2026-04-02
solution

Prus-Oleksandr:


app/cinema/bar.py
```diff
@@ -0,0 +1,10 @@
+from app.people.customer import Customer
+
+
+class CinemaBar:
+    def __init__(self) -> None:
```
andriihrechko: Tip: No __init__ method is required here. Since this class only contains static methods, it functions as a namespace and does not need to be instantiated. Use the methods directly via the class name.

andriihrechko (approved):
Well done!

PR #2626, opened 2026-04-17
Develop

RWAdelP:


RWAdelP:
@mateacademy-ai-mentor 
* "In app/people/customer.py, the static method CinemaBar.sell_product is currently defined with parameters in the order (customer, product), but the task explicitly specifies CinemaBar.sell_product(product, customer)"
The CinemaBar.sell_product(product, customer) method is in app/cinema/bar.py
* "In app/main.py, the cinema_visit function parameters are ordered differently from the specification (you currently take (customers, hall_number, cleaner, movie) instead of (movie, customers, hall_number, cleaner)),"
image
Running pytest gives the following error:
image

Digging in the test_main.py, the method call is as follows (Line 66):

cinema_visit(customers, hall_number, cleaner, movie)

vadim-ilchenko (approved):

PR #2663, opened 2026-05-05
implement Cinema

Shongi:


Shongi:
@mateacademy-ai-mentor
Just a small observation.

"The CinemaBar.sell_product method is defined as an instance method with self parameter, but the task explicitly states it should be a static method. You need to: 1. Add @staticmethod decorator above the sell_product method in both app/cinema/bar.py and app/people/cinema_staff.py"

While the "sell_product" method was definitely wrong and needs to be static, I don't understand the mention of the "cinema_staff.py" file as that has nothing to do with the faulty method.

PR #2681, opened 2026-05-18
Implement cinema visit workflow and domain classes.

Petliuk:
Add cinema and people packages with the required classes and print-based behaviors so cinema_visit orchestrates bar sales, movie session, and hall cleaning as required by tests.


Losiev (approved):
Well done!👏

PR #2756, opened 2026-06-26
Created two packages with some modules. Imported all modules in main …

MaksymMurtazin:
…and defined cinema_visit method that manages the cinema.


vsmutok (approved):
Good Job!

